PR #22, opened 2014-09-26
WIP: Control breakup

### LINKED ISSUE #20: Break up WP_Post_Edit_Customize_Control into field-specific controls
Instead of having one monolithic control, there should be separate controls for:

* Title
* Content
* Excerpt
* Publish Date
* Slug
* Author
* Status
* Parent
* Menu order
* Page template
* Featured image
* Meta (custom fields)

Doing it in this way will help users opt-in to the fields they need, and it would make it easier to extend and add new controls more modularly for metaboxes.

We'll also need to break up the settings as well.

The gotcha here is the "Select post to edit" dropdown. In the way it is right now, it would be hard to manage the breakup into individual controls because everything is in one section. An alternative is to switch to using panels, so each post resides within a section in a Posts panel. This would require sections to be dynamically generated in the panel, but there is no JS API for this right now (see [#28709](https://core.trac.wordpress.org/ticket/28709)).

In the mean time, it is probably better to get rid of the "Select post to edit" control altogether, and automatically load up posts when the preview has them as the queried object.

westonruter:
Fixes #20

### DIFF
--- a/bin/.jshintrc
+++ b/bin/.jshintrc
@@ -3,9 +3,11 @@
 	"curly": true,
 	"eqeqeq": true,
 	"eqnull": true,
+	"es3": true,
 	"expr": true,
 	"immed": true,
 	"noarg": true,
+	"onevar": true,
 	"quotmark": "single",
 	"trailing": true,
 	"undef": true,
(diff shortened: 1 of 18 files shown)


westonruter:
Made obsolete by #30.

PR #55, opened 2016-03-10
Add keyboard shortcut to open and close editor.

rheinardkorf:
- Creates a new customizePostCtrlShiftE trigger for Customizer and for the
  editor.

- Binds the new trigger to open and close the editor when Ctrl+Shift+E (PC) or
  Cmd+Shift+E (Mac) is pressed.

### DIFF
--- a/js/customize-post-section.js
+++ b/js/customize-post-section.js
@@ -269,7 +269,7 @@
 
 				var isCtrl = false, // Status of control/command key
 					isShift = false, // Status of shift key
-					modifier = navigator.appVersion.indexOf( 'Mac' ) !== -1 ? 91 : 17, // use command (91) for Mac
+					modifier = navigator.appVersion.indexOf( 'Mac' ) !== -1 ? 91 : 17, // Use command (91) for Mac
 					customizePostShortcuts = customizePostShortcuts || {};
 
 				customizePostShortcuts.keysUp = function( e ) {
--- a/js/customize-post-section.js
+++ b/js/customize-post-section.js
@@ -142,6 +142,9 @@
 		addContentControl: function() {
 			var section = this, control, setting = api( section.id ), editor;
 
+			// Get shortcut label for Mac or PC
+			var shortcutLabel = navigator.appVersion.indexOf( 'Mac' ) !== - 1 ? api.Posts.data.l10n.keyboardToggle.mac : api.Posts.data.l10n.keyboardToggle.pc;
+
 			control = new api.controlConstructor.dynamic( section.id + '[post_content]', {
 				params: {
 					section: section.id,
@@ -157,6 +160,8 @@
 			} );
 			control.editorExpanded = new api.Value( false );
 			control.editorToggleExpandButton = $( '<button type="button" class="button"></button>' );
+			/** Create container for shortcut key display */
+			control.editorToggleExpandButtonShortcut = $( '<div style="display: inline-block; padding: 5px;">' + shortcutLabel + '</div>' );
 			control.updateEditorToggleExpandButtonLabel = function( expanded ) {
 				control.editorToggleExpandButton.text( expanded ? api.Posts.data.l10n.closeEditor : api.Posts.data.l10n.openEditor );
 			};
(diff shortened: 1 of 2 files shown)


js/customize-post-section.js
```diff
@@ -227,9 +232,21 @@
...
+							editor.focus();
+						}
+					});
+
 				} else {
 					api.Posts.postIdInput.val( '' );
 					control.editorExpanded.set( false );
+
+					/** Remove Ctrl/Cmd+Shift+E trigger handler. */
+					$( document ).off( 'customizePostCtrlShiftE' );
```
westonruter: Oh, woops. I didn't notice this one! So yeah, my above comment is not actually true.

js/customize-post-section.js
```diff
@@ -243,6 +260,45 @@
...
+				customizePostShortcuts.keysUp = function( e ) {
+					isShift = 16 === e.which ? false : isCtrl;
+					isCtrl = e.which === modifier ? false : isCtrl;
+				};
+
+				customizePostShortcuts.keysDown = function( e ) {
+					isShift = 16 === e.which ? true : isCtrl;
+					isCtrl = e.which === modifier ? true : isCtrl;
+					if ( 69 === e.which && isCtrl && isShift ) {
+						$( document ).trigger( 'customizePostCtrlShiftE' );
```
westonruter: @rheinardkorf instead of triggering on the document, it would be better to use the `wp.customize` events for this. So you can do:

```js
wp.customize.trigger( 'toggle-post-editor' );
```

And then above you can listen for those events via:

```js
wp.customize.bind( 'toggle-post-editor', function() { /* ... */ } );
```

js/customize-post-section.js
```diff
@@ -243,6 +260,45 @@
...
+			/**
+			 * Create Ctrl/Cmnd+Shift+E shortcut trigger.
+			 * This adds the `customizePostCtrlShiftE` trigger to the document.
+			 */
+			( function() {
+
+				var isCtrl = false, // Status of control/command key
+					isShift = false, // Status of shift key
+					modifier = navigator.appVersion.indexOf( 'Mac' ) !== -1 ? 91 : 17, // Use command (91) for Mac
+					customizePostShortcuts = customizePostShortcuts || {};
```
westonruter: `customizePostShortcuts` is not defined above, so no need to `||` it here.

westonruter:
@rheinardkorf this isn't working in the context where there are more than one edit post in the preview (e.g. the homepage). It seems the event is getting triggered for each post even though they aren't expanded.

westonruter:
It seems I have to be focused in control input or the editor for the toggle to work. The toggle should work when focus is given to any element, including elements in the preview. For the preview, you can catch the keyboard events in `customize-preview-posts.js` and then `wp.customize.preview.send( 'toggle-post-editor' );` Then in `customize-post-section.js` you can listen for these messages and do the same thing as when the events happen: `wp.customize.previewer.bind( 'toggle-post-editor', function() { /* ... */ } )`

js/customize-post-section.js
```diff
@@ -243,6 +260,45 @@
...
+
+				customizePostShortcuts.keysUp = function( e ) {
+					isShift = 16 === e.which ? false : isCtrl;
+					isCtrl = e.which === modifier ? false : isCtrl;
+				};
+
+				customizePostShortcuts.keysDown = function( e ) {
```
westonruter: I think we should only need a `keyup` event handler, not also a `keydown` handler.

js/customize-post-section.js
```diff
@@ -243,6 +260,45 @@
...
+			( function() {
+
+				var isCtrl = false, // Status of control/command key
+					isShift = false, // Status of shift key
+					modifier = navigator.appVersion.indexOf( 'Mac' ) !== -1 ? 91 : 17, // Use command (91) for Mac
+					customizePostShortcuts = customizePostShortcuts || {};
+
+				customizePostShortcuts.keysUp = function( e ) {
+					isShift = 16 === e.which ? false : isCtrl;
+					isCtrl = e.which === modifier ? false : isCtrl;
```
westonruter: Modifying values outside the closure seems strange to me.

westonruter:
@rheinardkorf take a look at this modification on your approach: https://github.com/xwp/wp-customize-posts/commit/c9c78839fb9be74034ecddb4dbf2fead93bb4cb5

js/customize-post-section.js
rheinardkorf: To make it work better with wp triggering I will use a named function. That will make removing it better too, rather than accidentally removing someone else's handler.

westonruter:
@rheinardkorf cool, so how about cherry-picking that commit into your branch, and then the last outstanding bug I see is is the event listener doesn't work when the focus is *inside* the editor window. I think the problem is that there is some `keydown` event capturing that TinyMCE is doing which is preventing the event from bubbling up.

rheinardkorf:
@westonruter , happy for you to reject this PR now.  Just glad its on the cards now :D

rheinardkorf:
I see what you mean by the lost focus in the editor window.  It works for me if I'm using visual editor, but not if clicking on the kitchen sink... or if working in text editor.

westonruter:
Closing in a favor of intra-repo PR that @rheinardkorf can open via https://github.com/xwp/wp-customize-posts/compare/feature/editor-toggle-keyboard-shortcut?expand=1

PR #70, opened 2016-03-29
Implement post preview in Customizer

### LINKED ISSUE #3: Let “Preview” button in post edit open Customizer with post section autofocused
When previewing a post from the post edit screen, it should open in the Customizer. This avoids requiring the user to click an additional “Customize” link from the admin bar and to navigate to the post in the Customizer pane.

- [ ] ~~There needs to also be a post status dropdown in the Customizer (#40) so that drafts can actually be published from the Customizer.~~ (We can just skip allowing post to be published from Customizer view.)
- [x] When clicking “Preview” the Customizer should be loaded via `customize-loader` and any changes made to the post should be synced back to the parent frame in case “Save & Publish” is not clicked.
- [x] Clicking “Close” should not prompt with AYS confirmation because the value should have been synced to the editor in the parent frame.

See also http://wordpress.org/plugins/customizer-everywhere/

westonruter:
See demo: https://youtu.be/Q62nav1k4gY

Fixes #3.

### DIFF
--- a/js/edit-post-preview-admin.js
+++ b/js/edit-post-preview-admin.js
@@ -53,6 +53,8 @@ var EditPostPreviewAdmin = (function( $ ) {
 		settings[ postSettingId ] = postSettingValue;
 		sessionStorage.setItem( 'previewedCustomizePostSettings[' + postId + ']', JSON.stringify( settings ) );
 
+		wp.customize.Loader.open( self.data.customize_url );
+
 		// Sync changes from the Customizer to the post input fields.
 		wp.customize.Loader.messenger.bind( 'customize-post-settings-data', function( data ) {
 			var editor;
@@ -65,8 +67,6 @@ var EditPostPreviewAdmin = (function( $ ) {
 				$( '#content' ).val( data[ postSettingId ].post_content ).trigger( 'change' );
 			}
 		} );
-
-		wp.customize.Loader.open( self.data.customize_url );
 
 		wp.customize.Loader.settings.browser.mobile = wasMobile;
 	};
--- a/js/edit-post-preview-admin.js
+++ b/js/edit-post-preview-admin.js
@@ -1,0 +1,102 @@
+/* global jQuery, _editPostPreviewAdminExports, JSON, tinymce */
+/* exported EditPostPreviewAdmin */
+var EditPostPreviewAdmin = (function( $ ) {
+	'use strict';
+
+	var self = {
+		data: {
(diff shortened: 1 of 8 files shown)


valendesigns:
@westonruter Why hide the save button?

westonruter:
@valendesigns two reasons:

1. There are other fields on the post edit screen, including custom metaboxes, that aren't presented in the customizer (yet). So hitting “Save & Publish” would need to submit the form in the post edit screen anyway to ensure that the post is saved as expected.
2. The “Save & Publish” button is misleading if you are only wanting to save a draft, and I didn't want to re-engineer the button to handle that.

lukecarbis:
@valendesigns Because the save button applies only when there are multiple things to customise. It's a top level control that will save everything, and make the changes live immediately.

Within the context of a Preview, there's no need to have it there. In fact, it only introduces confusion. What should it say? "Save and Publish"? "Save and Update"? Just "Save"? What does that do? What if it's scheduled? etc.

lukecarbis:
@westonruter I noticed that there's a slightly jarring animation jumping straight into the post edit from the preview. The user is first shown the "Posts" parent item in the Customiser, then the Customiser animates over to the current post.

Of course, it would be better if this was displayed first, without the transition. 

In action: https://cloudup.com/cDQu-M2tnLl

valendesigns:
Thanks for the explanation guys.

westonruter:
@lukecarbis yeah, this would be a core issue. When the Customizer is loaded with a `panel`, `section`, or `control` autofocused, the root panel momentarily appears before animating the focused element into view. This will be fixed with the resolution of [#32667](https://core.trac.wordpress.org/ticket/32667) in core and would be outside the scope of this PR.

js/edit-post-preview-admin.js
```diff
@@ -0,0 +1,102 @@
...
+
+	self.init = function() {
+		$( '#post-preview' )
+			.off( 'click.post-preview' )
+			.on( 'click.post-preview', self.onClickPreviewBtn );
+	};
+
+	self.onClickPreviewBtn = function( event ) {
+		var $btn = $( this ),
+			postId = $( '#post_ID' ).val(),
```
lukecarbis: Not necessarily changeworthy, just a note. I always prefer using spaces to indent alignment with `var`. If ever 1 tab width is != 4 spaces (like on Github where 1 tab == 8 spaces), the alignment is skewed.

PR #71, opened 2016-03-29
Add show_in_customizer option for register_post_type()

westonruter:
Let a post type by default only be displayed in the Customizer if `show_ui` and `publicly_queryable` are true. Let this be overridden if `show_in_customizer` is defined.


johnregan3:
@westonruter Awesome work!  Always educational. :+1:

PR #89, opened 2016-04-10
Add postmeta setting and framework

### LINKED ISSUE #1: Add postmeta setting and base control
A page template and featured image are both stored in postmeta. These should use a `WP_Customize_Postmeta_Setting` with unique UIs for a page dropdown and (re-used) image media selector control. These can serve as the basis for creating more complex controls that map to multiple postmeta, or a control which manipulates a non-scalar postmeta value (an array).

### LINKED ISSUE #62: Add control for selecting post author
Changing the author will likely require refreshing the entire content `part` partial, or the entire preview, since there is not a discrete filterable way of displaying the post's author. If there is a partial specific for a `post_author` then this can be used, but if not present then the part needs to be refreshed (see #36), with ultimate fallback on refreshing the entire page.

### LINKED ISSUE #85: Add page template setting and control
Depends on #1.

westonruter:
Still needed:

- [x] Demonstration of adding a control.
- [x] Add missing unit tests.
- [x] Sync page template selection when doing page preview.

Fixes #1. 
Fixes #85.
Fixes #62.


php/class-wp-customize-posts.php
```diff
@@ -121,9 +141,110 @@ public function set_builtin_post_type_descriptions() {
...
+	 */
+	public function register_meta() {
+
+		foreach ( get_post_types( array(), 'objects' ) as $post_type_object ) {
+
+			if ( post_type_supports( $post_type_object->name, 'thumbnail' ) ) {
+				register_meta( 'post', '_thumbnail_id', array( $this, 'sanitize_post_id' ) );
+				$this->register_post_type_meta( $post_type_object->name, '_thumbnail_id' );
+			}
+		}
```
westonruter: This can be removed in favor of adding a controller in another PR.

js/customize-page-template.js
```diff
@@ -0,0 +1,102 @@
...
+				}
+			}
+		} );
+
+		/**
+		 * Make sure that control only appears if there are page templates (other than 'default').
+		 *
+		 * @returns {boolean}
+		 */
+		control.active.validate = function() {
```
valendesigns: This is having no effect on the controls visibility. I can see it when there is only the default template available.

php/class-customize-posts-plugin.php
```diff
@@ -169,6 +169,12 @@ public function register_scripts( WP_Scripts $wp_scripts ) {
 		$deps = array( 'customize-controls' );
 		$in_footer = 1;
 		$wp_scripts->add( $handle, $src, $deps, $this->version, $in_footer );
+
+		$handle = 'customize-page-template';
```
valendesigns: This resource could be tested in `Test_Customize_Posts_Plugin::test_register_scripts` for being registered and then tested for enqueued in the test class for `WP_Customize_Page_Template_Controller` that doesn't yet exist.

php/class-wp-customize-page-template-controller.php
```diff
@@ -0,0 +1,113 @@
+<?php
+/**
```
valendesigns: This docblock is a bit off.

php/class-wp-customize-page-template-controller.php
```diff
@@ -0,0 +1,113 @@
...
+
+	/**
+	 * Enqueue scripts.
+	 */
+	public function enqueue_scripts() {
+		$handle = 'customize-page-template';
+		wp_enqueue_script( $handle );
+		wp_add_inline_script( $handle, 'CustomizePageTemplate.init()' );
+
+		$choices = array();
```
valendesigns: Should the choices logic be moved to a method?

php/class-wp-customize-postmeta-controller.php
```diff
@@ -0,0 +1,147 @@
...
+		add_action( 'customize_controls_enqueue_scripts', array( $this, 'enqueue_scripts' ) );
+	}
+
+	/**
+	 * Register meta.
+	 */
+	public function register_meta() {
+
+		// Short-circuit if theme support is not present.
+		if ( isset( $this->theme_supports ) && ! current_theme_supports( $this->theme_supports ) ) {
```
valendesigns: Maybe it should be:
`! isset( $this->theme_supports ) || ! current_theme_supports( $this->theme_supports )`

westonruter: No, as we want to continue registering it if the `theme_supports` is empty, meaning it doesn't depend on a theme supporting a feature.

php/class-wp-customize-page-template-controller.php
westonruter: Done in bfa277d1fd1a8c0210f0ed7d94eb18fcefb4945d.

php/class-wp-customize-page-template-controller.php
westonruter: Done in dd4f7f792ff78ad7aafde4646d529112e1a9d28c.

js/customize-page-template.js
westonruter: Fixed in a0b6645a2e1e6bfe5f54e0b466d0aa3329aea248.

php/class-wp-customize-postmeta-setting.php
```diff
@@ -0,0 +1,206 @@
+<?php
+/**
```
valendesigns: Docblock has the wrong class name.

php/class-wp-customize-postmeta-setting.php
```diff
@@ -0,0 +1,206 @@
...
+	 * Meta key.
+	 *
+	 * @access public
+	 * @var string
+	 */
+	public $meta_key;
+
+	/**
+	 * Posts component.
+	 *
```
valendesigns: Missing `@access public`

php/class-wp-customize-postmeta-setting.php
```diff
@@ -0,0 +1,206 @@
...
+
+	/**
+	 * Return a post's setting value.
+	 *
+	 * @return mixed Meta value.
+	 */
+	public function value() {
+		$meta_key = $this->meta_key;
+		$object_id = $this->post_id;
+		$single = true;
```
valendesigns: What about post meta that is not `single`?

php/class-wp-customize-posts-preview.php
```diff
@@ -80,13 +89,10 @@ public function preview_setup_postdata( WP_Post $post ) {
 		if ( $prevent_setup_postdata_recursion ) {
 			return;
 		}
-		if ( ! $this->component->current_user_can_edit_post( $post ) ) {
```
valendesigns: Why remove the edit post check, did it get moved and checked elsewhere?

php/class-wp-customize-posts.php
```diff
@@ -121,9 +141,102 @@ public function set_builtin_post_type_descriptions() {
...
+				'transport' => null,
+				'sanitize_callback' => null,
+				'sanitize_js_callback' => null,
+				'setting_class' => 'WP_Customize_Postmeta_Setting',
+			),
+			$setting_args
+		);
+
+		if ( ! has_filter( "auth_post_meta_{$meta_key}", array( $this, 'auth_post_meta_callback' ) ) ) {
+			add_filter( "auth_post_meta_{$meta_key}", array( $this, 'auth_post_meta_callback' ), 10, 6 );
```
valendesigns: There are only 4 arguments in the method below.

php/class-wp-customize-postmeta-setting.php
westonruter: Good question. First of all, Customizing postmeta that have multiple meta values that have the same meta key is not supported. This is because the array passed into the `get_post_meta` filter doesn't include the `meta_id` to differentiate between multiple postmeta that have the same key. This is partly for simplicity but also because there is no key postmeta I'm aware of that actually needs this capability (e.g. page template and thumbnail ID both are only single meta values), but it's also because technically settings can't be registered with `$prev_value` in the setting ID.

But here specifically, in the `value` method, we're only concerned with `$single` because we're just getting a single value to return from `get_post_meta()`.

php/class-wp-customize-posts-preview.php
westonruter: Yes. It is now part of `$setting->check_capabilities()` below.

php/class-wp-customize-posts.php
westonruter: True. But the filter is passing in 6 arguments regardless. So the method below will just ignore those two additional args. I don't think it really matters.

valendesigns: You don't have to call all 6 though, you could just make it call the 4 you need.

php/class-wp-customize-postmeta-setting.php
valendesigns: Done in 02693ec

valendesigns:
@weston I'm a little confused about how `WP_Customize_Page_Template_Controller` works. From what I can tell the sanitize functions are not being used for the setting. I came across `Test_Customize_Postmeta_Setting::test_sanitize_page_template_setting` and noticed that the only way to get the test to pass was to explicitly add the `sanitize_callback` param in the `args` array when instantiating `WP_Customize_Postmeta_Setting`. Below is a passing test, though to me this demonstrates that the `args` are not being filtered because this should happen automatically.

``` PHP
function test_sanitize_page_template_setting() {
	switch_theme( 'twentytwelve' );

	$post_id = $this->factory()->post->create( array( 'post_type' => 'post') );
	$meta_key = '_wp_page_template';
	register_meta( 'post', $meta_key, array( $this->plugin->page_template_controller, 'sanitize_value' ) );
	$setting_id = WP_Customize_Postmeta_Setting::get_post_meta_setting_id( get_post( $post_id ), $meta_key );
	$setting = new WP_Customize_Postmeta_Setting( $this->manager, $setting_id, array( 
		'sanitize_callback' => array( $this->plugin->page_template_controller, 'sanitize_setting' )
	) );

	$this->assertEquals( 'default', $setting->sanitize( 'default' ) );
	$this->assertNull( $setting->sanitize( 'bad-template.php' ) );
	$page_template = 'page-templates/front-page.php';
	$this->assertEquals( $page_template, $setting->sanitize( $page_template ) );
}
```

westonruter:
@valendesigns 

> I'm a little confused about how `WP_Customize_Page_Template_Controller` works

You can see committed right now that the test is skipped:

```php
$this->markTestSkipped( 'Class is gone: WP_Customize_Page_Template_Postmeta_Setting' );
```

So it wasn't up to date. 

Now, it is expected that it would fail without the `sanitize_callback` arg supplied. Why? Because this filter makes sure that the supplied template actually refers to an actual page template in the theme. So when `WP_Customize_Page_Template_Controller::sanitize_setting()` the method is applying, it is expecting that the sanitized value should be `null`. And this is what the assertion is checking. When the sanitize callback is not present, the unit test fails as expected:

> Failed asserting that 'bad-template.php' is null.

There are actually two separate sanitizers that can apply to a postmeta setting. There is the only sanitizer that is added via `register_meta()`:

```php
register_meta( 'post', $meta_key, array( $this->plugin->page_template_controller, 'sanitize_value' ) );
```

This sanitizer is called when `sanitize_meta()` is called, and this is called in `\WP_Customize_Postmeta_Setting::sanitize()`. This method also calls apply-filters on `customize_sanitize_{$this->id}` which is what calls the `sanitize_callback`.  Why do we have two separate sanitizers? Because when `sanitize_meta()` is called, it doesn't supply the context of the post for the postmeta being sanitized. When the Customizer setting sanitizer is called, it has access to the `WP_Customize_Setting` instance, and thus has access to the post ID and post type that form part of the setting ID. This will allow us to eventually use the setting sanitizer to enforce a very interesting, although infrequently-used feature: you can have different page templates made available to different pages via the `theme_page_templates` filter. And this is enforced now by passing in the `$post` object:

```php
$page_templates = wp_get_theme()->get_page_templates( $post );
```

So yeah, I think what you found is expected behavior!

php/class-wp-customize-posts-preview.php
```diff
@@ -127,6 +133,112 @@ public function filter_the_posts_to_add_dynamic_post_settings_and_preview( array
...
+
+			return $single ? $value : array( $value );
+		} else {
+
+			$is_recursing = true;
+			$meta_values = get_post_meta( $object_id, '', $single );
+			$is_recursing = false;
+
+			foreach ( $this->previewed_postmeta_settings[ $object_id ] as $postmeta_setting ) {
+				if ( ! array_key_exists( $post_values, $postmeta_setting->id ) || ! $postmeta_setting->check_capabilities() ) {
```
mehigh: @westonruter 
When testing I got this notice, and it was related to this line.

array(1) { ["postmeta[page][10][_wp_page_template]"]=> string(31) "template-page-sidebar-right.php" } string(37) "postmeta[page][10][_wp_page_template]"
Warning: array_key_exists() expects parameter 2 to be array, string given in /srv/www/strathcomcms.com/wp-content/plugins/wp-customize-posts/php/class-wp-customize-posts-preview.php on line 228

I checked http://php.net/manual/ro/function.array-key-exists.php
and this was mentioning to turn those things - they key first, the array after, like from this:
array_key_exists( $post_values, $postmeta_setting->id )
into this:
array_key_exists( $postmeta_setting->id, $post_values )

js/customize-page-template.js
```diff
@@ -0,0 +1,113 @@
...
+
+		settingId = 'postmeta[' + section.params.post_type + '][' + String( section.params.post_id ) + '][_wp_page_template]';
+		controlId = settingId;
+
+		// Send the updated page template to the post edit screen when it is changed.
+		api( settingId, function( setting ) {
+			setting.bind( function( pageTemplate ) {
+				var settings = {};
+				settings[ settingId ] = pageTemplate;
+				EditPostPreviewCustomize.sendSettingsToEditPostScreen( settings );
```
mehigh: @westonruter 
This EditPostPreviewCustomize global yielded an invalid reference. I've updated Wordpress to 4.5 in our test environment and didn't get an improvement.
I've searched for this variable and since it's not in the current branch of the repository I'm not sure where it originates from. 
I've just commented this out (locally) and that yielded a fully functional page template adjustment without JS errors.

mehigh:
I had brought this into the strathcom project, since it worked. We'll be coming back if we find anything not to be working correctly.

js/customize-page-template.js
westonruter: @mehigh good catch! I've fixed this in f17aecb20f2c8ca196ccb135ccfc7c8a349522a0. The problem was that `EditPostPreviewCustomize` is only available in the context of opening the Customizer when in a post/page preview.

westonruter:
@valendesigns this epic PR I believe is now ready to merge!

PR #95, opened 2016-04-20
Fix compatibility with Customize Snapshots

westonruter:
Customize Posts is not working with Customize Snapshots.

I found that post and postmeta changes were not applied on the frontend when the UUID was supplied. This was due to the preview filters being added in the `customize_preview_init` action as opposed to being added whenever a setting's `preview()` method was called.

I also found that even with this fixed, the settings would not apply on the frontend for users who were not authenticated. There were a lot of capability checks added for _previewing_ settings which I now believe to be an error.

### DIFF
--- a/php/class-wp-customize-post-setting.php
+++ b/php/class-wp-customize-post-setting.php
@@ -479,7 +479,11 @@ class WP_Customize_Post_Setting extends WP_Customize_Setting {
 	 * @return bool
 	 */
 	public function preview() {
+		if ( $this->is_previewed ) {
+			return true;
+		}
 		$this->posts_component->preview->previewed_post_settings[ $this->post_id ] = $this;
+		$this->posts_component->preview->add_preview_filters();
 		$this->is_previewed = true;
 		return true;
 	}
--- a/php/class-wp-customize-post-setting.php
+++ b/php/class-wp-customize-post-setting.php
@@ -134,9 +134,6 @@ class WP_Customize_Post_Setting extends WP_Customize_Setting {
 	 * @return bool False if the post data did not apply.
 	 */
 	public function override_post_data( WP_Post &$post ) {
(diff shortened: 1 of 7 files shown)


westonruter:
@valendesigns thoughts on this?

valendesigns:
@westonruter Reviewing it this afternoon, sorry for the delay.

valendesigns:
@westonruter One thing that I noticed, and is technically unrelated from this PR, is that you cannot take a snapshot when you click "Preview Changes" because the button is gone. If we remove `#customize-header-actions #snapshot-save` from the CSS it comes back, what was your original reason for removing it? Otherwise, this PR looks good and can be merged. I'm just thinking it would be nice to return the "Save" button for snapshots. Thoughts?

westonruter:
@valendesigns I think that having a *Save* button when doing a post preview would be confusing. Users would surely think that the button here means to publish the post, not create a snapshot. If the button were renamed *Snapshot* that could make it more clear, as noted in https://github.com/xwp/wp-customize-snapshots/issues/13

And the Customizer post preview interface is mainly intended to be an abbreviated experience, as all of the other panels, sections, and controls are also hidden. So I think it makes sense to also hide the snapshot functionality, at least for now.

Regarding this PR, I'm mostly concerned about the removal of these capability checks and what that could mean for any malicious users. Could there be any vector to exploit?

valendesigns:
This is restricted to preview actions and is being set with `WP_Customize_Setting::post_value` which does sanitization on the data. Furthermore, capability checks are done in `WP_Customize_Setting::save` which would close exploit vectors from unauthenticated malicious users. I don't see any holes in security here. However, if there was one, then Snapshots would be more at risk than Customize Posts.

As for the button, we can leave it for now. 👍

PR #102, opened 2016-04-23
Add support for featured image

### LINKED ISSUE #57: Add featured image control and preview
Depends on #1.

westonruter:
Fixes #57.

- [x] Create a new setting representing the `_thumbnail_id` postmeta.
- [x] Create partial for rendering a featured image.
- [x] Add a `post_thumbnail_html` filter to inject the necessary class name for the partial `selector` to match.

### DIFF
--- a/php/class-wp-customize-featured-image-controller.php
+++ b/php/class-wp-customize-featured-image-controller.php
@@ -41,7 +41,7 @@ class WP_Customize_Featured_Image_Controller extends WP_Customize_Postmeta_Contr
 	 *
 	 * @var string
 	 */
-	public $setting_transport = 'refresh';
+	public $setting_transport = 'postMessage';
 
 	/**
 	 * Default value.
@@ -58,6 +58,7 @@ class WP_Customize_Featured_Image_Controller extends WP_Customize_Postmeta_Contr
 	public function __construct( array $args = array() ) {
 		parent::__construct( $args );
 		$this->override_default_edit_post_screen_functionality();
+		add_action( 'customize_register', array( $this, 'setup_selective_refresh' ) );
 	}
 
 	/**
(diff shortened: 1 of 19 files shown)


westonruter:
@valendesigns over to you, please review.

php/class-wp-customize-featured-image-controller.php
```diff
@@ -0,0 +1,376 @@
...
+	 *
+	 * @see sanitize_meta()
+	 *
+	 * @param int $attachment_id Attachment ID.
+	 * @return int|false Attachment ID, -1, or false if failure.
+	 */
+	public function sanitize_value( $attachment_id ) {
+		$attachment_id = intval( $attachment_id );
+		if ( empty( $attachment_id ) || -1 === $attachment_id ) {
+			return -1;
```
westonruter: This is wrong. It is resulting in `has_post_thumbnail()` always returning `true` because it casts the value to a `bool`. It needs to be an empty string instead. Other work needs to be done to fix this.

valendesigns:
@westonruter When you remove the featured image the anchor remains in the markup in the archive. As well, the article still has the class `.has-post-thumbnail` for both archive and single post. Leaving the class and markup in tact will result in UI issues for themes that style based on the image existing in the markup, like Twenty Fifteen for example.

screen shot 2016-04-25 at 3 39 36 pm

westonruter:
@valendesigns yes, I noticed this as well. I am adding a custom FeaturedImage partial which forces a refresh when the value is emptied, which should resolve this issue.

westonruter:
@valendesigns try now.

valendesigns:
@westonruter Yes, that change works perfectly. 👍

westonruter:
@valendesigns I think it's good to merge if you are happy.

php/class-wp-customize-post-field-partial.php
```diff
@@ -165,13 +165,20 @@ public function render_callback( WP_Customize_Partial $partial, $context = array
...
+					'author-bio',
+				);
+				foreach ( $template_parts as $template_part ) {
+					if ( '' !== locate_template( $template_part . '.php' ) ) {
+						ob_start();
+						get_template_part( $template_part );
+						$rendered = ob_get_contents();
+						ob_end_clean();
+						break;
+					}
```
westonruter: This chunk here relates to #103.

valendesigns: Shouldn't we do a check in the loop like

```PHP
if ( ! empty( $rendered ) ) {
    continue
}
```

westonruter: I don't think so. At this point we know the template exists, and if the template didn't render something, it's the theme's problem. In any case, all of this logic needs to be removed from this file and get defined in the theme instead, per #103.

PR #104, opened 2016-04-25
Add z-index for mce-floatpanel and mce-tooltip

stuartshields:
@westonruter I know it's bad to use !important, however in this case doing so resolves a lot of issues that are encountered. Also note that we cannot assume the end user wouldn't have plugins like TinyMCE Advanced installed. I've installed that particular plugin and tested the below and it works 100% without any issues. 

I'm open to alternative ways to achieve the end goal of the code below.

### DIFF
--- a/css/customize-posts.css
+++ b/css/customize-posts.css
@@ -76,14 +76,8 @@ body.mce-fullscreen.customize-posts-content-editor-pane-open div.mce-fullscreen 
 
 /* @todo Mobile support for rich text editor */
 
-.wp-customizer #wp-link-backdrop {
-	z-index: 500100; /* originally 100100, but z-index of .wp-full-overlay is 500000 */
-}
 .wp-customizer .ui-autocomplete.wplink-autocomplete {
 	z-index: 500110; /* originally 100110, but z-index of .wp-full-overlay is 500000 */
-}
-.wp-customizer div.mce-inline-toolbar-grp {
-	z-index: 500100; /* originally 100100, but z-index of .wp-full-overlay is 500000 */
 }
 .wp-customizer #wp-link-wrap {
 	z-index: 500105; /* originally 100105, but z-index of .wp-full-overlay is 500000 */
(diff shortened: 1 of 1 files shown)


css/customize-posts.css
```diff
@@ -76,15 +76,9 @@ body.mce-fullscreen.customize-posts-content-editor-pane-open div.mce-fullscreen
 
 /* @todo Mobile support for rich text editor */
 
-.wp-customizer #wp-link-backdrop {
```
stuartshields: @westonruter We can remove this as it is now covered by targeting .mce-panel. 

This applies also applies to the one below as well.

valendesigns:
@stuartshields Couldn't we use more specificity instead of using `!important`?

stuartshields:
@valendesigns I'll look more into this tonight. When I initially did it I was using more specificity however it caused a few issues in not working correctly. I'll report back what I find out so we can discuss more.

stuartshields:
@valendesigns I had a good look into this last night, it appears the way that WordPress brings in each element for the WYSIWYG editor in causes issues when you don't use `!important`, the code isn't generated on page load but only on editor load but below the editor generated element (if that makes sense). 

I've even attempted to use wild card (even though it's not really wild card) CSS to see if we could get this to work without `!important` without any luck.

Do you have any thoughts on the above? Any ideas that I might not have tried yet?

valendesigns:
@stuartshields Could you take a screen shot of the the issue so I can see what it is we're trying to fix in this PR?

westonruter:
@valendesigns it's the tooltips on the TinyMCE buttons.

stuartshields:
@valendesigns Tool tips, font colour drop down, Shortcut lightbox, font style dropdown.

Then you have anything that is generated by plugins like Advanced TinyMCE.

valendesigns:
It appears the only way to fix this without using `!important` is to do it in JS. I don't think it's worth the effort in this situation so I'm OK with the workaround. In the future we should probably patch the tooltip JS in Core if we can though.

PR #106, opened 2016-04-26
Define text_domain for i18n sniff

westonruter:
Depends on https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/560

With this we can get rid of `grunt-checktextdomain`.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -55,38 +55,6 @@ module.exports = function( grunt ) {
 			}
 		},
 
-		// Check textdomain errors.
-		checktextdomain: {
-			options:{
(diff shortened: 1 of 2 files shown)


valendesigns:
@westonruter So this would run on the pre-commit hook instead of with Grunt?

westonruter:
@valendesigns exactly.

valendesigns:
@westonruter Wouldn't we be removing functionality in the interim while PR#560 is being considered though?

westonruter:
Right, that's why we wait for that PR to be merged 😄

PR #107, opened 2016-04-27
Fix JS error

### LINKED ISSUE #45: Improve TinyMCE editor in mobile
The editor is not currently optimized for mobile in the Customize Posts integration.

Deeper UX/UI work is needed on the mobile editor, but some issues that could improve thing:

- [x] Collapse button for editor should not get overlapped by editor.
- [x] Editor shouldn't get positioned above other controls when pane is expanded. Editor should appear inline.
customize__wordpress_develop_ _just_another_wordpress_site_
- [x] Editor should be able to expand to full screen in mobile properly.
customize__wordpress_develop_ _just_another_wordpress_site_

### LINKED ISSUE #90: The editor cannot switch to Text from Visual mode

valendesigns:
By adding `autosave` as a dependency to the `customize-posts` JS the editor tabs and media button becomes functional again.

This suppresses `customize.php?return=%2Fwp-admin%2Fprofile.php:4243Uncaught TypeError: Cannot read property 'interval' of undefined`

I've also added some styles to improve the editor in mobile.

Fixes #90, and fixes #45

### DIFF
--- a/php/class-customize-posts-plugin.php
+++ b/php/class-customize-posts-plugin.php
@@ -168,7 +168,6 @@ class Customize_Posts_Plugin {
 			'customize-post-section',
 			'customize-dynamic-control',
 			'underscore',
-			'autosave',
 		);
 		if ( version_compare( str_replace( array( '-src' ), '', $GLOBALS['wp_version'] ), '4.6-beta1', '<' ) ) {
 			$deps[] = 'customize-controls-patched-36521';
--- a/php/class-customize-posts-plugin.php
+++ b/php/class-customize-posts-plugin.php
@@ -168,6 +168,7 @@ class Customize_Posts_Plugin {
 			'customize-post-section',
 			'customize-dynamic-control',
 			'underscore',
+			'autosave',
 		);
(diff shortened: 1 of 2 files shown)


php/class-customize-posts-plugin.php
```diff
@@ -168,6 +168,7 @@ public function register_scripts( WP_Scripts $wp_scripts ) {
 			'customize-post-section',
 			'customize-dynamic-control',
 			'underscore',
+			'autosave',
```
westonruter: I'd want to find out what the impact is of `autosave` being included in the Customizer.

valendesigns: We could also use `heartbeat` as the dependency. That seems like it would have less impact.

valendesigns: @westonruter The issue is that the following code is being added to the preview, which I thought was being added by Core. Come to find out it's from a MU plugin that was testing Snapshots. I'll revert the enqueue.

```PHP
<script>
	wp.customize.bind( 'ready', function() {
		wp.heartbeat.interval( 'fast' );
	} );
</script>
```

PR #111, opened 2016-04-27
Fix featured image in WP 4.5

### LINKED ISSUE #108: After Featured image is added in Customizer, error displays in WP Admin Featured image
**Test Steps**
Navigate to `{removed link}` and login
Click on Pages then >Add new page - WP only - Featured image
View Featured image box
_**Expected:**_
The featured image interface in WP Admin displays without error and works to upload a featured image

**Actual:**
On navigating to a new page, (even prior to saving the page) the Featured image box displays with error message - Warning: Missing argument 3 for WP_Customize_Featured_Image_Controller::filter_admin_post_thumbnail_html() in /srv/bindings/327e7c8b5e2b477ba002a93d24b46535/code/wp-content/plugins/customize-posts/php/class-wp-customize-featured-image-controller.php on line 189

newpagefeaturedimageerror

NB: the featured image functionality was working in WP Admin, prior to the customizer update to add the featured image in customizer.

valendesigns:
Fixes #108

### DIFF
--- a/js/customize-preview-featured-image.js
+++ b/js/customize-preview-featured-image.js
@@ -35,7 +35,7 @@ var CustomizePreviewFeaturedImage = (function( api, $ ) {
 		refresh: function() {
 			var partial = this, setting, refreshPromise;
 			setting = api( partial.params.primarySetting );
-			if ( '' === setting() ) {
+			if ( '' === setting() || -1 === setting() ) {
 				refreshPromise = $.Deferred();
 				partial.fallback();
 				refreshPromise.reject();
--- a/js/customize-preview-featured-image.js
+++ b/js/customize-preview-featured-image.js
@@ -35,7 +35,7 @@ var CustomizePreviewFeaturedImage = (function( api, $ ) {
 		refresh: function() {
 			var partial = this, setting, refreshPromise;
 			setting = api( partial.params.primarySetting );
-			if ( '' === setting() || -1 === setting() ) {
+			if ( '' === setting() ) {
 				refreshPromise = $.Deferred();
 				partial.fallback();
 				refreshPromise.reject();
--- a/php/class-wp-customize-featured-image-controller.php
+++ b/php/class-wp-customize-featured-image-controller.php
@@ -184,10 +184,16 @@ class WP_Customize_Featured_Image_Controller extends WP_Customize_Postmeta_Contr
 	 *
 	 * Executed during the `admin_post_thumbnail_size` hook to supply the current thumbnail ID
 	 * in WordPress < 4.6 where the third parameter in `admin_post_thumbnail_html` does not exist.
+	 *
+	 * @param string|array $size         Post thumbnail image size to display in the meta box.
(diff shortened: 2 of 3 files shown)


php/class-wp-customize-featured-image-controller.php
```diff
@@ -186,7 +206,10 @@ public function handle_save_post_thumbnail_id( $post_id ) {
 	 * @param int    $thumbnail_id Thumbnail ID.
 	 * @return string Content.
 	 */
-	public function filter_admin_post_thumbnail_html( $content, $post_id, $thumbnail_id ) {
+	public function filter_admin_post_thumbnail_html( $content, $post_id, $thumbnail_id = null ) {
+		if ( is_null( $thumbnail_id ) ) {
+			$thumbnail_id = $this->thumbnail_id;
```
westonruter: Would this be better as just `$thumbnail_id = get_post_thumbnail_id( $post_id );` instead of adding the `set_admin_post_thumbnail_id` function above?

php/class-wp-customize-featured-image-controller.php
```diff
@@ -171,6 +180,17 @@ public function handle_save_post_thumbnail_id( $post_id ) {
...
 
 	/**
+	 * Sets the current thumbnail ID.
+	 *
+	 * Executed during the `admin_post_thumbnail_size` hook to supply the current thumbnail ID
+	 * in WordPress < 4.6 where the third parameter in `admin_post_thumbnail_html` does not exist.
+	 */
+	public function set_admin_post_thumbnail_id( $size, $thumbnail_id, $post ) {
+		unset( $size, $post );
+		$this->thumbnail_id = $thumbnail_id;
```
westonruter: If this filter is to be used, it needs to return the `$size` here. It is getting lost otherwise.

js/customize-preview-featured-image.js
```diff
@@ -35,7 +35,7 @@ var CustomizePreviewFeaturedImage = (function( api, $ ) {
 		refresh: function() {
 			var partial = this, setting, refreshPromise;
 			setting = api( partial.params.primarySetting );
-			if ( '' === setting() ) {
+			if ( '' === setting() || -1 === setting() ) {
```
westonruter: See my comment here: https://github.com/xwp/wp-customize-posts/issues/108#issuecomment-215110422

php/class-wp-customize-featured-image-controller.php
valendesigns: That doesn't work unfortunately. When you add the featured image via the edit post page the hidden input does not get the correct ID set in WP 4.5.

php/class-wp-customize-featured-image-controller.php
valendesigns: Good catch, fixing it now.

php/class-wp-customize-featured-image-controller.php
westonruter: I see, because we're blocking the attachment ID from being saved, and so `get_post_thumbnail_id()` won't return the right thing until the post has been saved.

PR #116, opened 2016-04-28
Add avatar post field partial placements

westonruter:
Fixes issue in Twenty Sixteen with one of the avatars not getting updated.

avatar-not-updated

### DIFF
--- a/js/customize-post-field-partial.js
+++ b/js/customize-post-field-partial.js
@@ -55,9 +55,9 @@
 					if ( 'author-bio' === args.params.placement ) {
 						args.params.selector = baseSelector + ' .author-info';
 					} else if ( 'byline' === args.params.placement ) {
-						args.params.selector = baseSelector + ' .byline a.fn';
+						args.params.selector = baseSelector + ' .vcard a.fn';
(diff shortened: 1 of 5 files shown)


js/customize-post-field-partial.js
```diff
@@ -56,6 +56,8 @@
 						args.params.selector = baseSelector + ' .author-info';
 					} else if ( 'byline' === args.params.placement ) {
 						args.params.selector = baseSelector + ' .byline a.fn';
+					} else if ( 'avatar' === args.params.placement ) {
+						args.params.selector = baseSelector + ' img.avatar';
```
valendesigns: Maybe this should be `args.params.selector = baseSelector + ' .byline img.avatar';` incase there are other avatars in the post content?

westonruter: If there are other avatars in the post content (the post template part) the this is what we want. We won't be dependent on having a container that has a `byline` class.

valendesigns: I mean if there are other avatars that are not the author of the post/page.

PR #121, opened 2016-04-28
Update coveralls badge branch

westonruter:
See https://github.com/xwp/wp-dev-lib/pull/172


valendesigns:
Now that we have sorted the badge situation, do we really need this change? Seems like `master` should still be the branch for coverage.

westonruter:
I guess not.

PR #122, opened 2016-04-29
Fix the assets during deploy

valendesigns:


.dev-lib
```diff
@@ -1 +1,2 @@
 WPCS_GIT_TREE=develop
+ASSETS_DIR=wp-assets
```
valendesigns: Good catch!

PR #123, opened 2016-04-29
Add compatibility with themes and plugins

### LINKED ISSUE #82: Jetpack Support: Hide Feedback Posts in Customizer
When the Contact Form is enabled, the Feedback post type is registered and appears in the Customizer. Now with #71 we can hide the post type in the Customizer specifically, so we should add a Jetpack support class to the plugin and set this `show_in_customizer` arg to `false` for the `feedback` post type.

### LINKED ISSUE #103: Facilitate themes to configure partials
Instead of a big switch statement with hard-coded selectors and template parts, there needs to be more extensibility so that a theme can indicate how a partial should be rendered, rather than us assuming how it will be rendered. As part of this, we may want to have a collection of theme-specific code that gets loaded when a theme is active. In other words, we can bundle support for the partials unique to Twenty Fifteen, and provide examples for how themes should add support for selective refresh.

valendesigns:
Fixes #82
Fixes #103


westonruter:
@valendesigns I think we should have a separate class for each theme/plugin adding compatibility for. Eventually the theme/plugin should copy the compatibility into their own codebases, and then we can just delete our classes.

valendesigns:
@westonruter We could do it that way, sure. One thing that is not exactly clear though with partials is how to extend the JS and replace the selector for themes. Should we use an event or is there a way to override the selector in a simpler way?

westonruter:
@valendesigns That's a good question. I think that we can have a registry of the partials defined in PHP, with the selectors and render callbacks for each. And then the theme can filter this array. Or rather, it can be perhaps a theme support feature. That being the case, we can actually register the partials with PHP statically instead of dynamically with JS, or there can be a hybrid where theme partial configuration is fed into the filter for dynamic partials but the selectors and configurations (e.g. `fallback_refresh`, `container_inclusive`) get exported to JS for dynamic instantiation. (And yes, we actually do need to do dynamic instantiation to properly handle posts that get added dynamically to the document, such as for infinite scroll.)

php/class-customize-posts-plugin.php
```diff
@@ -57,6 +57,22 @@ public function __construct() {
...
 		}
 
+		// Theme & Plugin Support.
+		require_once dirname( __FILE__ ) . '/class-customize-posts-support.php';
+		require_once dirname( __FILE__ ) . '/class-customize-posts-plugin-support.php';
+		require_once dirname( __FILE__ ) . '/class-customize-posts-theme-support.php';
+
+		foreach ( array( 'theme', 'plugin' ) as $type ) {
+			foreach ( glob( dirname( __FILE__ ) . '/' . $type . '-support/class-*.php' ) as $file_path ) {
+				if ( is_readable( $file_path ) ) {
```
westonruter: I think the `is_readable()` is unnecessary?

php/class-customize-posts-support.php
```diff
@@ -0,0 +1,65 @@
...
+	 */
+	public $plugin;
+
+	/**
+	 * Plugin/Theme slug.
+	 *
+	 * @access public
+	 * @var string
+	 */
+	public $slug;
```
westonruter: Move this to `Customize_Posts_Plugin_Support`.

westonruter: Nevermind 😊

php/plugin-support/class-customize-posts-jetpack-support.php
```diff
@@ -0,0 +1,43 @@
...
+	 */
+	public $slug = 'jetpack/jetpack.php';
+
+	/**
+	 * Constructor.
+	 *
+	 * @access public
+	 */
+	public function add_support() {
+		add_filter( 'customize_posts_excluded_post_types', array( $this, 'excluded_post_types' ) );
```
westonruter: Instead of introducing a new filter, why not just assign it on the object itself?

```php
add_action( 'wp_loaded', function() {
    $obj = get_post_type_object( 'feedback' );
    if ( $obj ) {
        $obj->show_in_customizer = false;
    }
} );
```

valendesigns: We can't use closures, but this technique could work instead a using a filter.

westonruter: That's right. I just meant my code example as a proof of concept.

js/customize-preview-posts.js
```diff
@@ -39,23 +40,26 @@
 				// Post field partial for post_title.
 				partial = new api.previewPosts.PostFieldPartial( id + '[post_title]', {
 					params: {
-						settings: [ id ]
+						settings: [ id ],
+						selector: api.previewPosts.data.partialSelectors.title || '.entry-title'
```
westonruter: Good that you added a fallback selector. But I think it has to include the `baseSelector` here as well.

php/class-wp-customize-post-field-partial.php
```diff
@@ -111,46 +111,133 @@ public function render_callback( WP_Customize_Partial $partial, $context = array
...
+		/**
+		 * Filter the rendered partial.
+		 *
+		 * @param string|null          $rendered Rendered partial.
+		 * @param WP_Customize_Partial $partial  Partial.
+		 * @param array                $context  Context.
+		 * @param WP_Post              $post     Post object.
+		 * @return string|null
+		 */
+		$rendered = apply_filters( 'customize_posts_rendered_partial', $rendered, $partial, $context, $post );
```
westonruter: @valendesigns I don't think we need this filter. There is already `customize_partial_render` which can be used. See https://github.com/xwp/wordpress-develop/blob/018f988654bf0db14d83e075b3e850e919f95af0/src/wp-includes/customize/class-wp-customize-partial.php#L231-L255

js/customize-preview-posts.js
valendesigns: I can move it from `customize-post-field-partial.js` to here.

php/plugin-support/class-customize-posts-jetpack-support.php
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	/**
+	 * Exclude Jetpack registered post types from being displayed in the Customizer.
+	 *
+	 * @access public
+	 */
+	public function show_in_customizer() {
+		if ( Jetpack::is_module_active( 'contact-form' ) ) {
+			get_post_type_object( 'feedback' )->show_in_customizer = false;
```
westonruter: For good measure, I'd probably make sure that the `feedback` is actually registered. For example:

```php
if ( ! Jetpack::is_module_active( 'contact-form' ) ) {
    return;
}
$post_type_object = get_post_type_object( 'feedback' );
if ( ! $post_type_object ) {
    return;
}
$post_type_object->show_in_customizer = false;
```

php/class-wp-customize-post-field-partial.php
valendesigns: The only thing there would be that the filter is like `customize_partial_render_{$partial->id}` but `id` is like `post[:post_type][:post_id][:field_id]` and would be strange to filter on.

php/plugin-support/class-customize-posts-jetpack-support.php
valendesigns: Seems sane 👍

js/customize-preview-posts.js
```diff
@@ -123,6 +135,7 @@
 				partial = new api.previewPosts.PostFieldPartial( id + '[post_author][avatar]', {
 					params: {
 						settings: [ id ],
+						selector: api.previewPosts.data.partialSelectors.author.avatar || '.vcard img.avatar',
```
westonruter: For each of the dynamic partials being registered, there needs to be checks to see if the `partialSelectors` object actually exists: `api.previewPosts.data.partialSelectors.author.avatar`. Consider, for example, themes that don't make use of a given template part.

php/class-wp-customize-post-field-partial.php
valendesigns: Never mind, I misunderstood. You were only suggesting we remove the last filter.

php/class-wp-customize-posts-preview.php
```diff
@@ -442,11 +442,34 @@ public function export_preview_data() {
...
+				'avatar' => '.vcard img.avatar',
+			),
+		) );
+
 		$exported = array(
 			'isPostPreview' => is_preview(),
 			'isSingular' => is_singular(),
 			'queriedPostId' => $queried_post_id,
 			'settingProperties' => $setting_properties,
+			'partialSelectors' => $partial_selectors,
```
westonruter: Instead of exporting just the selectors, how about exporting the entire configs for the partials? For example, it could export here the `container_inclusive`, `fallback_refresh`, and other properties that will be used when the partials are registered on the client. This data can also then be extracted from being calculated in this method to instead be in a dedicated method, and it can then be used in `\WP_Customize_Posts_Preview::filter_customize_dynamic_partial_args()` instead of hard-coding those in `WP_Customize_Post_Field_Partial`: https://github.com/xwp/wp-customize-posts/blob/9ccc8f23a1589fc8098cac791c857781a605ef45/php/class-wp-customize-post-field-partial.php#L87-L94

php/class-wp-customize-post-field-partial.php
westonruter: That's right. `customize_posts_rendered_partial` is redundant and is already in core.

valendesigns: Should we change the other filter to be `"customize_partial_render_{$partial->field_id}"` to follow the same pattern `wp-customize-partial.php`?

valendesigns: Or `"customize_posts_partial_render_{$partial->field_id}"`?

westonruter:
@valendesigns one more thought: instead of having a default list of the post field partials that a theme is expected to support and the selectors for each, should these configurations not actually be added to `Customize_Posts_Theme_Support` subclasses for each Core theme?

php/class-wp-customize-post-field-partial.php
westonruter: Why do we need a filter here? It's not being used in this PR, and it seems the theme should be configuring these partials on the _input_ side of things (the partial configs) rather than the _output_, as presumably they would have supplied `render_callback` args that properly generate the right markup as-is.

valendesigns:
@weston That is a thought I had, but wanted to get some more input from you before going much further. with this PR. Essentially we need to come up with the best way to pass the partial config to the JS so it can be be added dynamically. You touched on that in one of the comments above, though that means we are relying on the support class to tell us which partials are supported. Which makes complete sense, we just need a good reusable schema to make this work well.

valendesigns:
@westonruter So we need to let the support class dictate the config, which is exported to JS and used to set the `$args` in `WP_Customize_Post_Field_Partial:__constructor`. The question then is how should the support class override the config?

valendesigns:
Also if the theme has not set a config then does that mean we should not register any partials?

westonruter:
@valendesigns

> So we need to let the support class dictate the config, which is exported to JS and used to set the `$args` in `WP_Customize_Post_Field_Partial:__constructor`. The question then is how should the support class override the config?

There can be a method like `WP_Customize_Posts::get_post_field_partial_configs()` which can have a filtered response. The theme supports can add filters for this response.

> Also if the theme has not set a config then does that mean we should not register any partials?

If the theme doesn't have a config set for partials, then we can just register the defaults which are supported by 99% of themes. For example any partials that target `the_content()` via an `.entry-content` selector: anything that makes use of the Microformat classes or anything output by `post_class()` should be safe to use as defaults.

valendesigns:
@westonruter I added the defaults which themes can unregister via the filter or modify as they see fit. I feel the defaults are fairly standard across many of the Twenty Something themes, there will be minor tweaks I'm sure, but overall themes should be using these partials. Have you seen the latest commit?

php/class-wp-customize-posts-preview.php
```diff
@@ -419,6 +425,84 @@ function filter_get_avatar( $avatar, $id_or_email, $size, $default, $alt, $args
...
+				),
+			),
+			'ping_status' => array(
+				'selector' => '.comments-area',
+				'body_selector' => true,
+				'singular_only' => true,
+				'container_inclusive' => true,
+			),
+			'post_author' => array(
+				'biography' => array(
```
westonruter: Do you think `biography` is standard enough to be a default?

php/class-wp-customize-posts-preview.php
```diff
@@ -419,6 +425,84 @@ function filter_get_avatar( $avatar, $id_or_email, $size, $default, $alt, $args
...
+				'selector' => '.entry-content',
+			),
+			'post_excerpt' => array(
+				'selector' => '.entry-summary',
+			),
+			'comment_status' => array(
+				'comments-area' => array(
+					'selector' => '.comments-area',
+					'body_selector' => true,
+					'singular_only' => true,
```
westonruter: What are `body_selector` and `singular_only`? I don't see them referenced anywhere.

php/class-wp-customize-posts-preview.php
valendesigns: I suppose bio is not going to be in all themes, we can remove it from the default list.

php/class-wp-customize-posts-preview.php
```diff
@@ -419,6 +425,84 @@ function filter_get_avatar( $avatar, $id_or_email, $size, $default, $alt, $args
...
+					'container_inclusive' => true,
+				),
+				'comments-link' => array(
+					'selector' => '.comments-link',
+					'archive_only' => true,
```
westonruter: Also, what is `archive_only`?

php/class-wp-customize-posts-preview.php
```diff
@@ -419,6 +425,84 @@ function filter_get_avatar( $avatar, $id_or_email, $size, $default, $alt, $args
...
+			),
+		);
+
+		/**
+		 * Filter the schema for dynamically registered partials.
+		 *
+		 * @param array $schema Partial schema.
+		 * @return array
+		 */
+		$schema = apply_filters( 'customize_posts_partial_schema', $schema );
```
westonruter: 👍

php/class-wp-customize-posts-preview.php
valendesigns: `body_selector` tells the JS to link the partials base selector to the body instead of the article, `singular_only` is to only load the partial on single posts, `archive_only` loads the partial everywhere but the single post. We can rework these, but they are needed to make sense of the context of a partial.

php/class-wp-customize-posts-preview.php
```diff
@@ -419,6 +425,84 @@ function filter_get_avatar( $avatar, $id_or_email, $size, $default, $alt, $args
...
+				'byline' => array(
+					'selector' => '.vcard a.fn',
+					'container_inclusive' => true,
+					'fallback_refresh' => false,
+				),
+				'avatar' => array(
+					'selector' => '.vcard img.avatar',
+					'container_inclusive' => true,
+					'fallback_refresh' => false,
+				),
```
westonruter: I liked the flatter schema format you had with `[post_author][avatar]` as keys, although I suppose the brackets are unsightly. Was that not used because of the trouble parsing out the parts? What if a theme wants to use a placement called `selector`? This would break the `previewPosts.partialSchema` logic for detecting nested structures.

What about using a key structure like `post_author[avatar]`? This would be easy to parse, as it's just removing all `]` and splitting on `[`.

php/class-wp-customize-posts-preview.php
westonruter: I see. I was looking for `body_selector` not `bodySelector` as it is named in JS.

php/class-wp-customize-posts-preview.php
valendesigns: We could check that `selector` is `typeof` string which would solve that?

westonruter: True. But is it better than having a flatter structure?

valendesigns: A flatter structure is nice sure, but then you couldn't so this in the filter...

```
$schema['post_author']['biography'] = array(
	'selector' => '.author-info',
	'singular_only' => true,
	'container_inclusive' => true,
);
```
You would be doing this instead...

```
$schema['[post_author][biography]'] = array(
	'selector' => '.author-info',
	'singular_only' => true,
	'container_inclusive' => true,
);
```

Which in my experience I believe will only confuse the masses.

westonruter: Or:

```php
$schema['post_author[biography]'] = array(
    'selector' => '.author-info',
    'singular_only' => true,
    'container_inclusive' => true,
);
```

Having multidimensional IDs are already very commonplace in the Customizer, e.g. `widget_text[123]` and `colors[background]`. So to me it seems having a multidimensional ID for the partials schema fits nicely.

valendesigns: I'll change it then 👍

php/theme-support/class-customize-posts-twenty-ten-support.php
```diff
@@ -0,0 +1,103 @@
...
+ * Add Customize Posts support.
+ *
+ * @codeCoverageIgnore
+ */
+function twentyten_support() {
+	add_theme_support( 'customize-posts' );
+	add_customize_posts_support( 'Customize_Posts_Twenty_Ten_Support' );
+}
+add_action( 'after_setup_theme', 'twentyten_support' );
+
```
westonruter: I think this might be a better approach:

```php
add_action( 'customize_register', function ( $wp_customize ) {
    if ( isset( $wp_customize->posts ) ) {
        $wp_customize->posts->add_support( 'Customize_Posts_Twenty_Ten_Support' );
    }
});
```

or

```php
add_action( 'customize_register', function ( $wp_customize ) {
    if ( isset( $wp_customize->posts ) ) {
        $wp_customize->posts->add_support( new Customize_Posts_Twenty_Ten_Support( $wp_customize->posts ) );
    }
});
```

customize-posts.php
```diff
@@ -31,3 +31,43 @@
...
+
+	if ( class_exists( $class_name, false ) ) {
+		$support = new $class_name();
+
+		if ( $support instanceof Customize_Posts_Support ) {
+			$support->init();
+			$wp_customize_posts_support[ $class_name ] = $support;
+		}
+	}
+}
```
westonruter: I think we can have `$wp_customize->posts->add_support()` which adds the instance to `$wp_customize->posts->supports` instead of a global `$wp_customize_posts_support`.

PR #133, opened 2016-05-10
Use `plugins_url()` directly

danielbachhuber:
This permits filtering of the final URL, which lets me use this plugin as a Git submodule without defining `SCRIPT_DEBUG`

### DIFF
--- a/php/class-customize-posts-plugin.php
+++ b/php/class-customize-posts-plugin.php
@@ -132,34 +132,33 @@ class Customize_Posts_Plugin {
 	 */
 	public function register_scripts( WP_Scripts $wp_scripts ) {
 		$suffix = ( SCRIPT_DEBUG ? '' : '.min' ) . '.js';
-		$plugin_dir_url = plugin_dir_url( dirname( __FILE__ ) );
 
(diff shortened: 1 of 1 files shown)


westonruter:
@danielbachhuber why not use Composer with WordPress Packagist instead? This would then include compiled and minified JS. Eventually if we use Browserify, using a submodule won't be an option.

danielbachhuber:
> why not use Composer with WordPress Packagist instead?

Because I'm managing dependencies with Git submodules. Also, I like the option to be able to easily contribute back while I'm working on the project.

westonruter:
OK, cool. But then eventually you'd need to run `grunt build` anyway if we do any Browserify compilation, right?

danielbachhuber:
> But then eventually you'd need to run grunt build anyway if we do any Browserify compilation, right?

Well, this would break my use of the project completely. My preference would be that you always commit a built copy of the assets to the repo.

westonruter:
@danielbachhuber You may not need to filter your script and style URLs anymore due to https://github.com/xwp/wp-customize-posts/pull/249

danielbachhuber:
Cool, thanks for the heads up 👍

PR #134, opened 2016-05-11
Post Creation

### LINKED ISSUE #48: Allow new posts and pages to be created
In addition to editing existing posts, entirely new posts/pages should also be able to be created in the Customizer. There are two approaches we could go here: 

(1) It could use the approach of creating new `nav_menu_item` posts where the pre-created post has a negative post ID which is used to designate a post that has not been inserted yet. 
(2) Or it could actually create an `auto-draft` post in the DB which would give us a real ID to reference.

It would definitely be less complicated to implement (2), since all the DB queries would work in the preview without having to inject pseudo-posts into query results. However, I also like (1) because it is more philosophically pure in that it truly does not touch the DB before Save & Publish.

Once the technical architecture for representing a pre-created post is settled, there is then the question of the UI for creating these posts. One integration option is in Customizer Nav Menus via the Add Item UI (see #49). Another integration option is to extend the post type's panel to add a “New Post” button before the list of post sections (see #50).

### LINKED ISSUE #50: Incorporate “Add New” buttons into post type panels
When a post type panel is expanded, before the list of posts currently shown in the preview, there should be a button to start authoring a new post of that type.

Depends on #48.

valendesigns:
Fixes #48, fixes #50

- [x] Create a new post/page and reload the preview while giving focus to that section
- [x] Ensure any newly created post/page persist in the panel
- [x] Add a `customize-draft` status which is assigned to an `auto-draft` post once it has been saved as part of a snapshot. This will prevent it from being garbage-collected.
- [x] Add link in a post section to allow navigating back to the post/page in the preview after having navigated away.
- [x] Display Ajax error messages in the console
- [x] Add unit tests
- [x] Rework the dependency on refreshing the preview to see new posts in the panel


westonruter:
@valendesigns what were the todos we listed out for this PR? I added one in the description. I realize we can continue to use `auto-draft` for posts added in the Customizer; we only need to introduce the `customize-draft` to prevent posts from being garbage-collected after they have become part of a snapshot. We'll need to think this through further with transactions, however.

valendesigns:
I'll update this later tonight I just got back from the airport. 😀

js/customize-posts-panel.js
```diff
@@ -64,6 +66,62 @@
...
+			addNewButton = wp.template( 'customize-posts-add-new' );
+			postObj = api.Posts.data.postTypes[ panel.postType ];
+
+			if ( postObj.current_user_can.create_posts ) {
+				descriptionContainer.after( addNewButton( {
+					label: postObj.labels.singular_name,
+					panel: panel
+				} ) );
+
+				$( '.add-new-' + panel.postType ).on( 'click', function( event ) {
```
westonruter: I believe this should be scoped, for example:

```js
panel.container.find( '.add-new.button-secondary' ).on( … )
```

In this case, there wouldn't need to be a type-specific class name because the button would be scoped to the container.

js/customize-posts-panel.js
```diff
@@ -64,6 +66,62 @@
 		},
 
 		/**
+		 * Add new post UI & listen for click events.
+		 */
+		addNewPost: function() {
```
westonruter: Let's rename this to something like `setupPostAddition`. A method like `addNewPost` would seem like a method we'd want to use to actually add a new post.

php/class-wp-customize-posts.php
```diff
@@ -77,6 +84,13 @@ public function __construct( WP_Customize_Manager $manager ) {
 		add_filter( 'customize_dynamic_setting_args', array( $this, 'filter_customize_dynamic_setting_args' ), 10, 2 );
 		add_filter( 'customize_dynamic_setting_class', array( $this, 'filter_customize_dynamic_setting_class' ), 5, 3 );
 		add_filter( 'customize_save_response', array( $this, 'filter_customize_save_response_for_conflicts' ), 10, 2 );
+		add_action( 'customize_controls_print_footer_scripts', array( $this, 'render_templates' ) );
+		add_action( 'init', array( $this, 'register_customize_draft' ) );
+		add_filter( 'customize_snapshot_save', array( $this, 'transition_customize_draft' ) );
+		add_action( 'pre_get_posts', array( $this, 'exclude_non_previewed_drafts' ) );
```
westonruter: I think this action should be added in `WP_Customize_Post_Setting::preview()` if not `has_action()` already. Similar to https://github.com/xwp/wordpress-develop/blob/8345fc60c7498b914d618b1b43a281fe269a0757/src/wp-includes/customize/class-wp-customize-nav-menu-item-setting.php#L396-L399

php/class-wp-customize-posts.php
```diff
@@ -567,4 +590,208 @@ public function sanitize_post_id( $value ) {
...
+		if ( ! $post_type_obj || ! current_user_can( $post_type_obj->cap->create_posts ) ) {
+			wp_send_json_error( 'insufficient_post_permissions' );
+		}
+
+		$post = $this->add_new_post( $_POST['post_type'] );
+
+		if ( $post instanceof WP_Post ) {
+			$data = array(
+				'sectionId' => WP_Customize_Post_Setting::get_post_setting_id( $post ),
+				'url' => get_preview_post_link( $post->ID ),
```
westonruter: Should this be using `Edit_Post_Preview::get_preview_post_link()` instead?

php/class-wp-customize-posts.php
```diff
@@ -567,4 +590,208 @@ public function sanitize_post_id( $value ) {
...
+			status_header( 403 );
+			wp_send_json_error( 'customize_not_allowed' );
+		}
+
+		if ( empty( $_POST['post_type'] ) ) {
+			status_header( 400 );
+			wp_send_json_error( 'missing_post_type' );
+		}
+
+		$post_type_obj = get_post_type_object( $_POST['post_type'] );
```
westonruter: `wp_unslash()` for good measure.

php/class-wp-customize-posts.php
```diff
@@ -567,4 +590,208 @@ public function sanitize_post_id( $value ) {
...
+		if ( ! $post_type_obj || ! current_user_can( $post_type_obj->cap->create_posts ) ) {
+			wp_send_json_error( 'insufficient_post_permissions' );
+		}
+
+		$post = $this->add_new_post( $_POST['post_type'] );
```
westonruter: Use `$post_type_object->name` instead of accessing input var again.

php/class-wp-customize-posts.php
```diff
@@ -567,4 +590,208 @@ public function sanitize_post_id( $value ) {
...
+	 * @access public
+	 *
+	 * @param string $post_type The post type.
+	 */
+	public function add_new_post( $post_type = '' ) {
+		if ( null !== get_post_type_object( $post_type ) ) {
+			add_filter( 'wp_insert_post_empty_content', '__return_false' );
+			$args = array(
+			  'post_status' => 'auto-draft',
+			  'post_type'   => $post_type,
```
westonruter: Lacking tabs for indenting the array args.

php/class-wp-customize-posts.php
```diff
@@ -567,4 +590,208 @@ public function sanitize_post_id( $value ) {
...
+
+		return $link;
+	}
+
+	/**
+	 * Add a new `auto-draft` post.
+	 *
+	 * @access public
+	 *
+	 * @param string $post_type The post type.
```
westonruter: Missing `@return`

php/class-wp-customize-posts.php
```diff
@@ -567,4 +590,208 @@ public function sanitize_post_id( $value ) {
...
+		return $permalink;
+	}
+
+	/**
+	 * Filter the preview permalink for a page.
+	 *
+	 * @access public
+	 *
+	 * @param string $link    The page's permalink.
+	 * @param int    $post_id The ID of the page.
```
westonruter: Missing `@return`

php/class-wp-customize-posts.php
```diff
@@ -567,4 +590,208 @@ public function sanitize_post_id( $value ) {
...
+		}
+	}
+
+	/**
+	 * Filter the preview permalink for a post.
+	 *
+	 * @access public
+	 *
+	 * @param string  $permalink The post's permalink.
+	 * @param WP_Post $post      The post in question.
```
westonruter: Missing `@return`

js/customize-posts-panel.js
```diff
@@ -64,6 +66,62 @@
...
+								completeCallback: function() {
+									setTimeout( callback, 500 );
+								}
+							} );
+						} );
+					} );
+
+					request.fail( function() {
+
+						// @todo Display errors in the Customize Settings Validation area.
```
westonruter: For now we can just `console.error()` it out. The notification area depends on https://core.trac.wordpress.org/ticket/35210

php/class-edit-post-preview.php
```diff
@@ -172,6 +184,7 @@ public function enqueue_customize_scripts() {
 	 */
 	public function make_auto_draft_status_previewable() {
 		global $wp_post_statuses;
+		$wp_post_statuses['auto-draft']->public = true;
 		$wp_post_statuses['auto-draft']->protected = true;
```
westonruter: Maybe this should also be added to `WP_Customize_Post_Setting::preview()` since `customize_preview_init` does not trigger in a frontend request?

valendesigns: We don't need `auto-draft` on the frontend because they are transitioned to `customize-draft` in snapshots.

php/class-wp-customize-posts.php
```diff
@@ -77,6 +84,13 @@ public function __construct( WP_Customize_Manager $manager ) {
 		add_filter( 'customize_dynamic_setting_args', array( $this, 'filter_customize_dynamic_setting_args' ), 10, 2 );
 		add_filter( 'customize_dynamic_setting_class', array( $this, 'filter_customize_dynamic_setting_class' ), 5, 3 );
 		add_filter( 'customize_save_response', array( $this, 'filter_customize_save_response_for_conflicts' ), 10, 2 );
+		add_action( 'customize_controls_print_footer_scripts', array( $this, 'render_templates' ) );
+		add_action( 'init', array( $this, 'register_customize_draft' ) );
```
westonruter: Should this also be added to `WP_Customize_Post_Setting::preview()`? Could there be an scenario where a non-`publish` post could get leaked accidentally otherwise? Or rather, maybe the preview method could grab the post status object and set the `public` property to `true`?

php/class-wp-customize-posts.php
```diff
@@ -567,4 +590,208 @@ public function sanitize_post_id( $value ) {
...
+
+					// @todo Amend customize_save_response.
+				} else {
+					$data[ $id ]['value']['post_status'] = $post_status;
+				}
+			}
+		}
+
+		return $data;
+	}
```
westonruter: Nice work on `transition_customize_draft`.

php/class-wp-customize-posts.php
```diff
@@ -567,4 +590,208 @@ public function sanitize_post_id( $value ) {
...
+	 *
+	 * @action pre_get_posts
+	 * @access public
+	 *
+	 * @param WP_Query $query The WP_Query instance (passed by reference).
+	 */
+	public function exclude_non_previewed_drafts( $query ) {
+		if ( $query->is_main_query() && ! is_admin() && ! $query->is_single() && ! empty( $_REQUEST['customized'] ) ) {
+			$post_ids = array();
+			$settings = json_decode( wp_unslash( $_REQUEST['customized'] ), true );
```
westonruter: Replace this (and `$_REQUEST['customized']` above) with `$settings = $this->manager->unsanitized_post_values()`.

php/class-wp-customize-posts.php
```diff
@@ -567,4 +590,208 @@ public function sanitize_post_id( $value ) {
...
+				}
+			}
+			$draft_query = new WP_Query( array(
+				'posts_per_page' => -1,
+				'post_status'    => self::$draft_status,
+				'fields'         => 'ids',
+			) );
+			$excluded = array_diff( $draft_query->posts, $post_ids );
+			if ( ! empty( $excluded ) ) {
+				$query->set( 'post__not_in', $excluded );
```
westonruter: This should _amend_ to any pre-existing `post__not_in`.

valendesigns: Makes sense!

php/class-wp-customize-posts.php
valendesigns: We sure can.

php/class-wp-customize-posts.php
```diff
@@ -567,4 +590,208 @@ public function sanitize_post_id( $value ) {
...
+				}
+				if ( in_array( $setting['post_status'], self::$draft_status, true ) ) {
+					$post_ids[] = intval( $matches['post_id'] );
+				}
+			}
+			$draft_query = new WP_Query( array(
+				'posts_per_page' => -1,
+				'post_status'    => self::$draft_status,
+				'fields'         => 'ids',
+			) );
```
westonruter: @valendesigns I wonder if there couldn't be a way to _include_ as opposed to _exclude_ posts? If I 

This query for getting _all_ posts could hit a lot of posts.

Let me confirm my understanding for how Customizer draft posts are appearing in the query: currently it is setting the post status for `auto-draft` and `customize-draft` to being `public` so that they will be included in the call to `get_post_stati()` to obtain the list of post statuses that should be used in a query: https://github.com/xwp/wordpress-develop/blob/8345fc60c7498b914d618b1b43a281fe269a0757/src/wp-includes/query.php#L3093-L3116

But what if the current `WP_Query` _did_ include a specified `post_status`? Even with `'post_status' => 'publish'`, which is the default arg supplied when calling `get_posts()`, it would result in these Customizer posts not being included. I suppose that's in part why you have the `$query->is_main_query()` check?

Here's an idea for an alternate approach for how we could perhaps go about being inclusive rather than exclusive:

1. Let our statuses not have the `public` post status.
2. Leave the original `pre_get_posts` alone to do its normal thing.
3. Add a `the_posts` filter which takes the current query vars for the current `$query` and then creates a new `WP_Query` that supplies the query vars merged with ``posts__in` => $post_ids`, to obtain all of the posts in the Customizer dirty state that align with the query vars in the existing query. (Naturally there'd need to be a mechanism to prevent infinite `the_posts` filter recursion here.)
4. The `the_posts` filter for previewing posts data should have overridden each of the posts at this point to use the actual `post_status` from the current Customizer state. 
5. The list of `$posts` in the subquery can now be iterated over and any can be rejected that have a previewed `post_status` that does not align with the `post_status` from the parent query.
6. The remaining posts can be amended to the post list that is returned by `the_posts`, ensuring that no duplicates are introduced, that the number returned does not exceed the `posts_per_page`, and that the posts are re-sorted according to `order` and `orderby`.

The result here should be that the query is more efficient, but also (more importantly) it will ensure that you will be take an existing `draft` or `private` post, change its status to `publish` _in the Customizer_, and be able to see such posts in queries for `publish` status posts in the preview. At the moment, the logic here is only accounting for posts that have never been saved.

Does this make sense?

js/customize-posts-panel.js
```diff
@@ -64,6 +66,62 @@
...
+					event.preventDefault();
+
+					request = wp.ajax.post( 'customize-posts-add-new', {
+						'customize-posts-nonce': api.Posts.data.nonce,
+						'wp_customize': 'on',
+						'post_type': panel.postType
+					} );
+
+					request.done( function( response ) {
+						wp.customize.previewer.previewUrl( response.url );
```
westonruter: @valendesigns in thinking about adding new Nav Menu Items, I'm second guessing delegating to the preview to create the new post section once the preview loads. I think that we should include the post setting data in the Ajax response and pass it along to `receivePreviewData` (but extracting the logic from there to be useful in scenarios where the preview refreshes _or_ if the post is added without intending to navigate to that post in the preview). 

This will also be key for adding new posts that are not `public` but rather are embedded within other posts.

php/class-wp-customize-posts.php
valendesigns: @westonruter I've figured out a way to be inclusive by filtering `posts_where`, and it's working with CPT's; which is pretty magical. I'm just writing the unit tests and I'll push this up. I think it makes for a very big improvement over the previously exclusive code and should be much more performant. As well, we no longer will be setting `public` to `true` for these draft statuses, which had some very strange side effects. Also, pagination is working without any special sorting.

westonruter: Looking forward to seeing the magic 🌟

js/customize-posts-panel.js
```diff
@@ -64,6 +66,64 @@
 		},
 
 		/**
+		 * Add new post stub, which builds the UI & listens for click events.
+		 */
+		setupPostAddition: function() {
```
westonruter: @valendesigns please break out the innards of this method into a new `wp.customize.Posts.insertPost()` method which takes as its argument an initial post object, e.g. `{ post_type: 'page', post_title: 'About Us' }`. This will be then used here and also in the available nav menu items interface that @celloexpressions has created the UI for in https://core.trac.wordpress.org/ticket/34923#comment:32

The `wp.customize.Posts.insertPost()` method can return a promise which resolves with the `wp.customize.Section` added as well. That would then allow us to do:

```js
wp.customize.Posts.insertPost( { post_type: 'page' } ).done( function( data ) {
    data.section.focus();
} );
```

valendesigns: @westonruter In order to pass an entire post object we would need to sanitize all the arguments. It would be much simpler to only allow the post title and post type, which is the basic information we need for nav menu items. Technically we only need the post type to stub a new post but the post title makes the setting dirty and will ensures the section doesn't disappear on a full refresh.

westonruter: @valendesigns yeah, we can handle passing the full post object pretty easily. Consider:

```php
<?php
$exported_settings = array();
$post_setting_id = WP_Customize_Post_Setting::get_post_setting_id( $auto_draft_post );
$wp_customize->add_dynamic_settings( array( $post_setting_id ) )

// BEGIN NEW STUFF
$wp_customize->set_post_value( $post_setting_id, $post_array_sent_in_ajax_request );
$post_setting = $wp_customize->get_setting( $post_setting_id );
$post_setting->preview();
// END NEW STUFF

$setting_ids = $wp_customize->posts->register_post_type_meta_settings( $new_post_id );
$setting_ids[] = $post_setting_id;
foreach ( $setting_ids as $setting_id ) {
	$setting = $wp_customize->get_setting( $setting_id );
	if ( ! $setting ) {
		continue;
	}
	if ( method_exists( $setting, 'json' ) ) { // new in 4.6-alpha
		$exported_settings[ $setting->id ] = $setting->json();
	} else {
		$exported_settings[ $setting->id ] = array(
			'value' => $setting->js_value(),
			'transport' => $setting->transport,
			'dirty' => $setting->dirty,
			'type' => $setting->type,
		);
	}
}
// Now send $exported_settings in Ajax response
```

js/customize-posts-panel.js
```diff
@@ -64,6 +66,42 @@
...
+
+			if ( postObj.current_user_can.create_posts ) {
+				descriptionContainer.after( addNewButton( {
+					label: postObj.labels.singular_name
+				} ) );
+
+				panel.container.find( '.add-new-post-stub' ).on( 'click', function( event ) {
+					event.preventDefault();
+
+					api.Posts.insertPost( { post_type: panel.postType }, true ).done( function( section ) {
```
westonruter: Instead of `true` for preview, perhaps we could just pass in `postObj.public`?

valendesigns:
@westonruter This PR is ready for a final review.

php/class-wp-customize-posts.php
```diff
@@ -583,4 +611,358 @@ public function sanitize_post_id( $value ) {
...
+	 * @access public
+	 *
+	 * @param string   $where The WHERE clause of the query.
+	 * @param WP_Query $query The WP_Query instance (passed by reference).
+	 * @return string
+	 */
+	public function include_previewed_drafts( $where, $query ) {
+		global $wpdb;
+
+		if ( ! $query->is_admin() && ! $query->is_singular() ) {
```
westonruter: PhpStorm is complaining here that `is_admin()` is not a method found on `WP_Query`. I can't find a `WP_Query::is_admin()` method either, but also it is not throwing an error for me which is also unexpected. I suppose it is because there is a magic getter defined.

So that leads me to think that `$query->is_admin()` is actually always returning `null` and isn't returning the `bool` that you expect if accessing `$query->is_admin`.

valendesigns: `is_admin()` will work.

valendesigns: We can change it to `if ( ! is_admin() && ! $query->is_singular() ) {`

westonruter: For some reason `$wp_query->is_admin()` is returning a bool. However note:

```
$query = new WP_Query();
$query->is_admin = true;
$this->assertTrue( $query->is_admin() );
```

This test fails. All of this to say, I think that `$query->is_admin` should be used instead of `$query->is_admin()`

valendesigns:
@westonruter Looks good!

php/class-wp-customize-posts.php
westonruter: My comment is no longer valid.

js/customize-posts-panel.js
```diff
@@ -64,9 +66,88 @@
...
+						 */
+						function focusControlOnceFocusable() {
+							var firstControl = data.section.controls()[0];
+							function onChangeActive( isActive ) {
+								if ( isActive ) {
+									data.section.active.unbind( onChangeActive );
+									_.defer( function() {
+										firstControl.focus( {
+											completeCallback: function() {
+												firstControl.container.find( 'input:first' ).select();
```
westonruter: Haha. I really did do some egregious callback nesting here didn't I 😆

php/class-wp-customize-posts-preview.php
```diff
@@ -165,6 +166,91 @@ public function filter_the_posts_to_preview_settings( array $posts ) {
...
+					continue;
+				}
+
+				/**
+				 * Filter the post type used in the main query.
+				 *
+				 * @param string $post_type         Main query post type.
+				 * @param string $setting_post_type Current setting post type.
+				 */
+				$main_query_post_type = apply_filters( 'customize_posts_main_query_post_type', 'post', $matches['post_type'] );
```
westonruter: I'm not totally sure about the need for this filter.

valendesigns: The idea was that if the main query is being filtered manually you could add those post types if needed, but probably don't need it.

westonruter: OK, done in https://github.com/xwp/wp-customize-posts/commit/9cdcb7c6594331f3298c81f9e963cdc43d5dac3d

Issue #34, opened 2016-03-04
Optionally filter edit_post_link on frontend to deep-link to post in Customizer


westonruter:

westonruter:
The `url` param should be added here as well and it should be the current URL on which the `edit_post_link()` appears, as opposed to the permalink for the post. This would allow the Customizer preview to initially load in the same view and when the link was clicked.

stuartshields:
@westonruter This doesn't appear to support custom post types... Is that correct?

Even though the URL is formatted correctly, it gives us a error 404.
http://vip.local/wp-admin/customize.php?url=http%3A%2F%2Fvip.local%2F%3Fpreview%3Dtrue%26p%3D22&amp;previewed_post=22&amp;autofocus[section]=post[movies][22]&amp;customize_preview_post_nonce=8745488e02

westonruter:
@stuartshields it does work with custom post types.

Your URL here looks like it has HTML encoding of entities.

westonruter:
Good on you for ensuring the url of the post being previewed is also included in the edit post link, _with_ the autofocus to open the section when the customizer loads. I didn't note that in the description.

stuartshields:
@westonruter I haven't done anything that was already in the code as references here: https://github.com/xwp/wp-customize-posts/blob/develop/php/class-edit-post-preview.php#L108

I don't like repeating myself so I'm thinking I'll build on what's already done. You can't see any issues with me doing this can you? I should have this and #93 ready by the end of the week. I also noticed when using this deep link with #93 that no Save/Publish button appears have you come across that issue?

westonruter:
@stuartshields yes, extracting that logic into a separate `get_customize_url` method is great.

>  I also noticed when using this deep link with #93 that no Save/Publish button appears have you come across that issue?

Yes. I see that too. This is due to `\Edit_Post_Preview::can_load_customize_post_preview()` incorrectly returning `true` in this instance. Basically, since you're re-using the same logic to generate the URL to access that post in the preview, the existing URL generated is expecting that you are loading it _from_ the edit post screen via clicking the Preview Changes button. When opening the Customizer via the Preview Changes button, it disables the Save & Publish button and removes all controls/sections/panels not relevant to the current post. This is undesirable when clicking Edit in Customizer generally. The quickest workaround is to prevent `previewed_post` from being added as a query var expect in the URL generated for clicking the Preview Changes button.

PR #140, opened 2016-05-18
Feature / Add edit in customizer buttons

stuartshields:
@westonruter and @valendesigns here is the initial concept for adding the `Edit in Customizer` button to the Posts/Pages edit screens. As well as adding `Edit in Customizer` to both `post_row_actions` and `page_row_actions`. I moved the `customize url` into it's own function since we're calling it in two different functions now.

### DIFF
--- a/php/class-edit-post-preview.php
+++ b/php/class-edit-post-preview.php
@@ -36,6 +36,8 @@ class Edit_Post_Preview {
 		add_action( 'customize_controls_init', array( $this, 'remove_static_controls_and_sections' ), 100 );
 		add_action( 'customize_controls_enqueue_scripts', array( $this, 'enqueue_customize_scripts' ) );
 		add_action( 'customize_preview_init', array( $this, 'make_auto_draft_status_previewable' ) );
+		add_filter( 'post_row_actions', array( $this, 'add_edit_customizer_to_row_actions' ), 10, 2 );
+		add_filter( 'page_row_actions', array( $this, 'add_edit_customizer_to_row_actions' ), 10, 2 );
 	}
 
 	/**
@@ -101,20 +103,54 @@ class Edit_Post_Preview {
 			return;
 		}
 		wp_enqueue_script( 'edit-post-preview-admin' );
+		$customize_url = self::get_customize_url();
 
 		$data = array(
-			'customize_url' => self::get_customize_url(),
+			'customize_url' => $customize_url,
 		);
+
 		wp_scripts()->add_data( 'edit-post-preview-admin', 'data', sprintf( 'var _editPostPreviewAdminExports = %s;', wp_json_encode( $data ) ) );
 		wp_enqueue_script( 'customize-loader' );
 		wp_add_inline_script( 'edit-post-preview-admin', 'jQuery( function() { EditPostPreviewAdmin.init(); } );', 'after' );
+
+		if ( 'add' !== get_current_screen()->action ) {
+			wp_add_inline_script( 'edit-post-preview-admin', 'jQuery( \'.wrap h1\' ).append( \'<a class="page-title-action hide-if-no-customize" href="'. esc_url( $customize_url ) .'">Edit in Customizer</a>\' )', 'after' );
+		}
 	}
 
 	/**
-	 * Get Post customize link
+	 * Add the edit customizer to row actions for Posts/Pages
+	 *
+	 * @param array $actions
+	 * @param object $post
+	 * @return array $rebuild_actions
 	 */
-	public function get_customize_url() {
-		$post = $this->get_previewed_post();
+	public function add_edit_customizer_to_row_actions( $actions, $post ) {
+		if ( ! is_a( $post, 'WP_Post' ) ) {
+			return false;
+		}
+
+		// Let's rebuild the order of $actions
+		$rebuild_actions = array();
+		$rebuild_actions['edit'] = $actions['edit'];
+		$rebuild_actions['edit_customizer'] = '<a href="' . esc_html( self::get_customize_url( $post ) ) . '">Edit in Customizer</a>';
+		$rebuild_actions['inline hide-if-no-js'] = $actions['inline hide-if-no-js'];
+		$rebuild_actions['trash'] = $actions['trash'];
+		$rebuild_actions['view'] = $actions['view'];
+
+		return $rebuild_actions;
+	}
+
+	/**
+	 * Get the customize link
+	 * 
+	 * @param array $post
+	 * @return string $customize_url
+	 */
+	public function get_customize_url( $post = null ) {
+		if ( ! is_a( $post, 'WP_Post' ) ) {
+			$post = $this->get_previewed_post();
+		}
 
 		$id_param = ( 'page' === $post->post_type ) ? 'page_id' : 'p';
 		$url = get_preview_post_link( $post, array(), home_url( '?preview=true&' . $id_param . '=' . $post->ID ) );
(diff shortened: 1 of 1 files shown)


php/class-edit-post-preview.php
```diff
@@ -101,7 +103,54 @@ public function enqueue_admin_scripts() {
...
+			return false;
+		}
+
+		// Let's rebuild the order of $actions.
+		$rebuild_actions = array();
+		$rebuild_actions['edit'] = $actions['edit'];
+		$rebuild_actions['edit_customizer'] = '<a href="' . esc_html( self::get_customize_url( $post ) ) . '">Edit in Customizer</a>';
+		$rebuild_actions['inline hide-if-no-js'] = $actions['inline hide-if-no-js'];
+		$rebuild_actions['trash'] = $actions['trash'];
+		$rebuild_actions['view'] = $actions['view'];
```
westonruter: What about other plugins that add action links? I think an `array_merge()` is required.

php/class-edit-post-preview.php
```diff
@@ -101,7 +103,54 @@ public function enqueue_admin_scripts() {
...
+	 */
+	public function add_edit_customizer_to_row_actions( $actions, $post ) {
+		if ( ! is_a( $post, 'WP_Post' ) ) {
+			return false;
+		}
+
+		// Let's rebuild the order of $actions.
+		$rebuild_actions = array();
+		$rebuild_actions['edit'] = $actions['edit'];
+		$rebuild_actions['edit_customizer'] = '<a href="' . esc_html( self::get_customize_url( $post ) ) . '">Edit in Customizer</a>';
```
westonruter: `esc_url()` instead of `esc_html()`

Also, use `esc_html__( 'Edit in Customizer', 'customize-posts' )`

php/class-edit-post-preview.php
```diff
@@ -36,6 +36,8 @@ public function __construct( Customize_Posts_Plugin $plugin ) {
 		add_action( 'customize_controls_init', array( $this, 'remove_static_controls_and_sections' ), 100 );
 		add_action( 'customize_controls_enqueue_scripts', array( $this, 'enqueue_customize_scripts' ) );
 		add_action( 'customize_preview_init', array( $this, 'make_auto_draft_status_previewable' ) );
+		add_filter( 'post_row_actions', array( $this, 'add_edit_customizer_to_row_actions' ), 10, 2 );
+		add_filter( 'page_row_actions', array( $this, 'add_edit_customizer_to_row_actions' ), 10, 2 );
```
westonruter: This should probably loop over all post types that are are eligible for showing in the Customizer instead of hard-coding just two. There is a method for getting all post types and code in the plugin for determining if the post type should `show_in_customizer`

php/class-edit-post-preview.php
```diff
@@ -101,7 +103,54 @@ public function enqueue_admin_scripts() {
...
+		$data = array(
+			'customize_url' => $customize_url,
+		);
+
+		wp_scripts()->add_data( 'edit-post-preview-admin', 'data', sprintf( 'var _editPostPreviewAdminExports = %s;', wp_json_encode( $data ) ) );
+		wp_enqueue_script( 'customize-loader' );
+		wp_add_inline_script( 'edit-post-preview-admin', 'jQuery( function() { EditPostPreviewAdmin.init(); } );', 'after' );
+
+		if ( 'add' !== get_current_screen()->action ) {
+			wp_add_inline_script( 'edit-post-preview-admin', 'jQuery( \'.wrap h1\' ).append( \'<a class="page-title-action hide-if-no-customize" href="' . esc_url( $customize_url ) . '">Edit in Customizer</a>\' )', 'after' );
```
westonruter: I'd suggest adding the link in the footer, and then use jQuery to move this existing link in the document into the right place, instead putting the markup into a JS string. 

Also needs translation function.

php/class-edit-post-preview.php
stuartshields: @westonruter definitely.

php/class-edit-post-preview.php
```diff
@@ -101,25 +105,83 @@ public function enqueue_admin_scripts() {
...
+
+	/**
+	 * Add the edit customizer to row actions for Posts/Pages.
+	 *
+	 * @param array  $actions Actions.
+	 * @param object $post Post.
+	 * @return array $rebuild_actions
+	 */
+	public function add_edit_customizer_to_row_actions( $actions, $post ) {
+		if ( ! is_a( $post, 'WP_Post' ) ) {
```
valendesigns: Let's be consistent and use the same coding style. Other places in the plugin we would check if `$post` is an instance of `WP_Post` like this:

```
if ( ! ( $post instanceof WP_Post ) ) {
	return false;
}
```

stuartshields: @valendesigns Thanks I've now fixed it.

php/class-edit-post-preview.php
```diff
@@ -101,25 +105,84 @@ public function enqueue_admin_scripts() {
...
+	}
+
+	/**
+	 * Get the customize line.
+	 *
+	 * @param array $post Post.
+	 * @return string $customize_url
+	 */
+	public function get_customize_url( $post = null ) {
+		if ( ! is_a( $post, 'WP_Post' ) ) {
```
valendesigns: This one could also be updated.

php/class-edit-post-preview.php
```diff
@@ -169,9 +169,10 @@ public function get_customize_url( $post = null ) {
 		if ( ! $post ) {
 			return false;
 		}
-		
+
 		$id_param = ( 'page' === $post->post_type ) ? 'page_id' : 'p';
-		$url = get_preview_post_link( $post, array(), home_url( '?preview=true&' . $id_param . '=' . $post->ID ) );
+
+		$url = get_preview_post_link( $post, array(), get_permalink( $post->ID ) . '?preview=true&' . $id_param . '=' . $post->ID );
```
stuartshields: @westonruter I've added get_permalink here as I can't work out how we use `get_edit_post_link`, so maybe some direction?

php/class-edit-post-preview.php
```diff
@@ -101,25 +105,84 @@ public function enqueue_admin_scripts() {
...
+		}
+
+		if ( ! $post ) {
+			return false;
+		}
 
 		$id_param = ( 'page' === $post->post_type ) ? 'page_id' : 'p';
-		$url = get_preview_post_link( $post, array(), home_url( '?preview=true&' . $id_param . '=' . $post->ID ) );
+
+		$url = get_preview_post_link( $post, array(), get_permalink( $post->ID ) . '?preview=true&' . $id_param . '=' . $post->ID );
```
valendesigns: You can get the URL with `Edit_Post_Preview::get_preview_post_link( $post )`

stuartshields:
@westonruter and @valendesigns This is ready for another look at. Any comments/direction would be good. As far as I can tell this is working the way we want it to, however I'm open to making this work better.

php/class-edit-post-preview.php
stuartshields: @valendesigns Thanks! I'll do this tonight :)

valendesigns:
@stuartshields It appears you could just do away with the `get_customize_url` method altogether and use the static `get_preview_post_link` instead.

stuartshields:
@valendesigns I wasn't able to get rid of get_customize_url altogether as it still needs the other variables that are generated to work. Latest commit has the fix, if you wanted to have a look.

valendesigns:
@stuartshields The post object is all you need to generate the preview URL so you can totally replace the `get_customize_url`. What else do see as a requirement that `get_preview_post_link` doesn't offer?

stuartshields:
@valendesigns I guess I'm concerned about not using `get_customize_url` because of #34  ... Do we continue with #34 but not use it for this one?

valendesigns:
@stuartshields I've updated the PR a bit and resolved conflicts. All that's left is to add unit test coverage for the new methods.

Issue #42, opened 2016-03-09
Turn off pretty permalinks for posts in the Customizer preview


westonruter:
This will help ensure that changes to the `post_name` won't result unexpectedly in the preview showing a 404, or requiring the preview to change the URL being previewed when editing a post.

PR #148, opened 2016-05-26
Add post slug field

### LINKED ISSUE #63: Add control for manipulating post slug
This will likely not have any impact on the preview. See #42.

westonruter:
customize_

Placeholder field updates live as title changes.

Fixes #63

### DIFF
--- a/php/class-wp-customize-posts.php
+++ b/php/class-wp-customize-posts.php
@@ -522,6 +522,7 @@ final class WP_Customize_Posts {
 				/* translators: &#9656; is the unicode right-pointing triangle, and %s is the section title in the Customizer */
 				'sectionCustomizeActionTpl' => __( 'Customizing &#9656; %s', 'customize-posts' ),
 				'fieldTitleLabel' => __( 'Title', 'customize-posts' ),
+				'fieldSlugLabel' => __( 'Slug', 'customize-posts' ),
 				'fieldContentLabel' => __( 'Content', 'customize-posts' ),
 				'fieldExcerptLabel' => __( 'Excerpt', 'customize-posts' ),
 				'fieldDiscussionLabel' => __( 'Discussion', 'customize-posts' ),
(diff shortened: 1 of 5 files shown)


valendesigns:
@westonruter We need to filter the post link so it uses the preview URL if the slug is dirty or we get a 404. I'm looking into it now.

westonruter:
@valendesigns Is this #42? I was thinking the changing of the slug was going to be a problem, but the full page refreshes continued to work even when the slug was being changed. I wasn't able to see a 404.

valendesigns:
@westonruter I can make it work without #42.

valendesigns:
@westonruter So it's not pretty but it does the job.

westonruter:
@valendesigns what were the steps that you used to reproduce the 404?

valendesigns:
@westonruter 
1) Turned on permalinks
2) Edited an existing post from the archive
3) Attempted to navigate to it which produced an 404

westonruter:
@valendesigns wouldn't it be better to just use preview links across the board? See 97ddef1.

valendesigns:
That sounds like a brilliant idea, I love it!

php/class-wp-customize-posts.php
```diff
@@ -702,58 +702,18 @@ public function transition_customize_draft( $data ) {
...
-			add_filter( 'post_link', array( $this, 'post_link_draft' ), 10, 2 );
-			add_filter( 'post_type_link', array( $this, 'post_link_draft' ), 10, 2 );
-
-			$post_status_match = in_array( get_post_status( $post->ID ), self::$draft_status, true );
-			$post_slug_mismatch = get_option( 'permalink_structure' ) && basename( $old_permalink ) !== $post->post_name;
-
-			if ( $post_status_match || $post_slug_mismatch ) {
-				$permalink = Edit_Post_Preview::get_preview_post_link( $post );
-			}
+		if ( is_customize_preview() ) {
```
valendesigns: We want this on the front-end for Snapshots, as well.

westonruter: @valendesigns Humm. Why should it be on the snapshotted frontend as well? 

And if so, what would be the best way to detect if it is a frontend snapshot preview?

valendesigns: @westonruter We can worry about it later, it's not critical since we haven't setup a way to navigate in the snapshot on the front-end. We'll tackle that later. For now we can just merge and move forward. Thoughts?

PR #150, opened 2016-05-26
Add support for setting validation in 4.6-alpha

### LINKED ISSUE #142: Add support for setting validation now in 4.6-alpha
Setting validation is now in `trunk`. See https://core.trac.wordpress.org/changeset/37476

The validation being done in Customize posts is using the proposed `$strict` param which didn't make it into the final patch. We need to implement `validate` callbacks, or return `WP_Error` from the settings' `sanitize` callbacks (but only if at least `4.6-alpha-r37476`, aka if `WP_Customize_Setting::validate()` exists).

westonruter:
image

Validation for post locking and post conflicts is only applied at save.

Post locking:

image

Post conflicts:

image

Fixes #142.

### DIFF
--- a/css/customize-posts.css
+++ b/css/customize-posts.css
@@ -1,7 +1,3 @@
-/* In case Customize Setting Validation is not active */
-div.customize-setting-validation-message.error {
-	display: none;
-}
 
 .customize-posts-panel-notice {
 	color: #555;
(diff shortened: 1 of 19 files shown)


valendesigns:
@westonruter I'm getting some weird errors with this PR

issues

valendesigns:
Glad we cleared that up, having the Customize Setting Validation plugin active and not the most recent version was the issue.

PR #152, opened 2016-05-28
Post Status Control

### LINKED ISSUE #40: Add post status dropdown selection control and preview
When opening a draft post in the Customizer, it would be ideal to be able to make changes and then _publish_ the post from there. For example in a workflow to preview how changes to a title wraps: https://youtu.be/sXu2pA42J88

As part of this, we can also implement the Trash button (#137) alongside the post status dropdown.

### LINKED ISSUE #137: Delete a post whilst in the Customizer
It would be great if I could trash or delete a post whilst in the Customizer.

I recognize that this task isn't small, and will need some UX exploration, etc. before it can go to development.

valendesigns:
Fixes #40, fixes #137

### DIFF
--- a/js/customize-post-section.js
+++ b/js/customize-post-section.js
@@ -321,9 +321,9 @@
 			} );
 
 			/**
-			 * Update the UI when a post is trasitioned from/to trash.
+			 * Update the UI when a post is transitioned from/to trash.
 			 *
-			 * @param {bool} trashed - Whether or not the post_status is 'trash'.
+			 * @param {boolean} trashed - Whether or not the post_status is 'trash'.
 			 * @returns {void}
 			 */
 			control.toggleTrash = function( trashed ) {
--- a/js/customize-post-section.js
+++ b/js/customize-post-section.js
@@ -368,10 +368,8 @@
 				}, embeddedDelay );
 			} );
 
-			// Remove the setting from the settingValidationMessages since it is not specific to this field.
-			if ( control.settingValidationMessages ) {
-				control.settingValidationMessages.remove( setting.id );
-				control.settingValidationMessages.add( control.id, new api.Value( '' ) );
+			if ( control.notifications ) {
+				control.notifications.add = section.addPostFieldControlNotification;
 			}
 			return control;
 		},
--- a/js/customize-post-section.js
+++ b/js/customize-post-section.js
@@ -179,9 +179,9 @@
 			if ( postTypeObj.supports.title || postTypeObj.supports.slug ) {
 				section.addSlugControl();
 			}
-
-			section.addPostStatusControl();
-
+			if ( 'undefined' === typeof EditPostPreviewCustomize ) {
+				section.addPostStatusControl();
+			}
 			if ( postTypeObj.supports.editor ) {
 				section.addContentControl();
 			}
--- a/js/customize-post-section.js
+++ b/js/customize-post-section.js
@@ -121,7 +121,7 @@
 			customizeActionElement = sectionInnerTitleElement.find( '.customize-action' ).first();
 			setting.bind( function( newPostData, oldPostData ) {
 				var title;
-				if ( newPostData.post_title !== oldPostData.post_title ) {
+				if ( newPostData.post_title !== oldPostData.post_title && 'trash' !== newPostData.post_status ) {
 					title = newPostData.post_title || api.Posts.data.l10n.noTitle;
 					sectionOuterTitleElement.text( title );
 					sectionInnerTitleElement.text( title );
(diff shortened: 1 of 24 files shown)


js/customize-preview-posts.js
```diff
@@ -150,6 +150,10 @@
 	};
 
 	api.bind( 'preview-ready', function() {
+		api.bind( 'saved', function() {
+			api.preview.send( 'customized-posts-purge-trash', api.previewPosts.data.isSingular );
+		} );
```
westonruter: @valendesigns I think this isn't right. The `saved` event is triggered on the `api` inside the controls itself. Why send this `customized-posts-purge-trash` message from the preview?

Why not just do in `customize-posts.js`:

```js
api.bind( 'saved', function() {
    component.purgeTrash();
} );
```

js/customize-post-section.js
```diff
@@ -179,6 +179,9 @@
 			if ( postTypeObj.supports.title || postTypeObj.supports.slug ) {
 				section.addSlugControl();
 			}
+
+			section.addPostStatusControl();
```
westonruter: @valendesigns since sync isn't supported yet, maybe we should just skip adding the `post_status` control altogether if `EditPostPreviewCustomize` is defined?

js/customize-post-section.js
```diff
@@ -281,6 +284,96 @@
...
+						'default': setting.id
+					},
+					field_type: 'select',
+					setting_property: 'post_status',
+					choices: api.Posts.data.postStatusChoices
+				}
+			} );
+
+			/**
+			 * Update the UI when a post is trasitioned from/to trash.
```
westonruter: Typo `s/trasitioned/transitioned/`

js/customize-preview-posts.js
valendesigns: `is_singular()` always returned `false` when called from the pane. So I needed access to that value to force the preview to reload the archive.

js/customize-post-section.js
```diff
@@ -281,6 +284,96 @@
...
+			 *
+			 * @param {bool} trashed - Whether or not the post_status is 'trash'.
```
westonruter: `s/bool/boolean/` PhpStorm likes `boolean` over `bool`. It looks like `bool` is just for PHP. It's not mentioned in JSDoc: http://usejsdoc.org/tags-type.html

js/customize-post-section.js
valendesigns: Sounds good. How do we access that from `customize-post-section.js`

js/customize-preview-posts.js
westonruter: I think it would be better to keep track of whatever the last data sent from `customized-posts` was instead.

```js
component.previewedQuery = new api.Value( {} );
api.previewer.bind( 'customized-posts', function ( data ) {
    var query = {};
    _.each( [ 'isSingular', 'isPostPreview', 'queriedPostId' ] ).each( function( key ) {
        query[ key ] = data[ key ];
    } );
    component.previewedQuery.set(  ) );    
} );
```

Then at any time you can access whether the preview `isSingular` via `query.previewedQuery.get().isSingular`.

This would be a basic implementation of https://core.trac.wordpress.org/ticket/36582

js/customize-post-section.js
westonruter: Just check if `typeof EditPostPreviewCustomize !== 'undefined'`.

php/class-wp-customize-post-setting.php
```diff
@@ -508,12 +503,23 @@ protected function update( $data ) {
...
-			return false;
+		if ( $is_trashed ) {
+			add_filter( 'wp_insert_post_empty_content', '__return_false', 100 );
 		}
-		return true;
+		$r = wp_update_post( wp_slash( $data ), true );
+		$result = is_wp_error( $r );
+
+		if ( $result && $is_trashed ) {
+			$result = wp_trash_post( $is_trashed );
```
westonruter: This needs to be:

```php
$result = wp_trash_post( $this->post_id );
```

westonruter:
- [ ] Add test to ensure a post with an empty title, content, and excerpt can be trashed.
- [ ] Add test to ensure inserted auto-draft post has the expected guid.
- [ ] Add test for `filter_customize_save_response_to_export_saved_values`.

php/class-wp-customize-post-setting.php
```diff
@@ -307,15 +306,15 @@ public function sanitize( $post_data, $strict = false ) {
 		$post_data = sanitize_post( $post_data, 'db' );
 		$initial_sanitized_post_data = $post_data;
 
-		$maybe_empty = 'attachment' !== $this->post_type
+		$maybe_empty = 'trash' !== $post_data['post_status'] && 'attachment' !== $this->post_type
```
westonruter: I think this is wrong. We should instead be doing this change below:

```php
if ( 'trash' !== $post_data['post_status'] && apply_filters( 'wp_insert_post_empty_content', $maybe_empty, $post_data ) ) {
```

Issue #172, opened 2016-06-09
Improve UX for deleting posts


fjarrett:
Thanks to #48 the user can now create new posts and pages. But what if they want to then discard a newly drafted post/page?

With the current implementation, in order to "undo" a newly created post/page the user must leave the Customizer and discard _all_ other changes that they might have staged in the session.

Additionally, it seems reasonable that the user should be able to trash posts/pages that are pre-existing, just as they can inside `post.php`.

![delete-trash-content](https://cloud.githubusercontent.com/assets/522158/15948734/1628ff98-2e68-11e6-8fa4-fcff84b42d78.png)

valendesigns:
The way that works is to set the post status to trash which will clean up any newly created or existing posts when the state is saved.

westonruter:
@fjarrett:

> But what if they want to then discard a newly drafted post/page?

Just change the Post Status to Trash?

> With the current implementation, in order to "undo" a newly created post/page the user must leave the Customizer and discard all other changes that they might have staged in the session.

Actually no, they just have to change the Post Status to Trash.

> Additionally, it seems reasonable that the user should be able to trash posts/pages that are pre-existing, just as they can inside post.php.

Yeah, this can be done by changing the Post Status to Trash:

<img width="291" alt="trash" src="https://cloud.githubusercontent.com/assets/134745/15948926/9d8fcea4-2e58-11e6-9496-89aba93117a0.png">

I think the confusion here may be that users don't expect “Trash” to be among the status dropdown. Maybe there should be a shortcut link like “Move to trash” that would select this status for them?

fjarrett:
@valendesigns OK got it. Though not very intuitive or consistent from a UI perspective.

<img width="300" alt="screen shot 2016-06-09 at 5 41 20 pm" src="https://cloud.githubusercontent.com/assets/522158/15948941/7c7bc61c-2e69-11e6-8442-2982aadc131b.png">

valendesigns:
It's not consistent but we can add the trash link to fix that which would select trash in the post status for now. We'll do more work on that control in the future. We still need to sync back to the admin and this would be a good time to address the big picture concerns.

fjarrett:
Cool, thanks for the feedback.

I just tested selecting the "Trash" post status on a newly-created page and clicking Save & Publish and it actually puts the page in the Trash. Which means it's still creating a post in the database.

So when a user clicks "Add New Post" they only have two options to back out:

[...]

PR #173, opened 2016-06-09
Delete auto-draft and customize-draft status posts when saving a trash post_status

westonruter:
See #172.


fjarrett:
@westonruter Tested and works perfectly!

Trashing a newly-staged post/page does not produce any database entry on Publish.

PR #180, opened 2016-06-14
Add support for setting-property control notifications and focusing

westonruter:
When a `post` setting is marked as invalid and the returned `WP_Error` includes a `setting_property` field, ensure that the corresponding control for this notification gets the notification displayed and gets focus when attempting to submit.

Depends on https://core.trac.wordpress.org/ticket/36944


westonruter:
@valendesigns the patch in wpcore#36944 has been committed so this can be tested against `trunk` and merged.

PR #191, opened 2016-06-28
Improve meta query support

PatelUtkarsh:
Add `IN` Meta query compare support. 

Using `$publish` flag to filter out posts which do not match meta query.

Add filter `customize_previewed_posts_for_query` for filtering previewed posts.

### DIFF
--- a/php/class-wp-customize-posts-preview.php
+++ b/php/class-wp-customize-posts-preview.php
@@ -367,6 +367,23 @@ final class WP_Customize_Posts_Preview {
 				}
 			}
 		}
+		/**
+		 * Filter customize preview posts.
+		 *
+		 * @param array $post_ids Post ids being filtered.
+		 * @param array array {
+		 *     Args.
+		 *
+		 *     @type \WP_Query $query   WP_Query obj.
+		 *     @type array     $settings Field Values in snapshot.
+		 *     @type bool      $publish IN query or NOT IN query.
+		 * }
+		 */
+		$post_ids = apply_filters( 'customize_previewed_posts', $post_ids, array(
+			'query' => $query,
+			'settings' => $settings,
+			'publish' => $published,
+		) );
 
 		return $post_ids;
 	}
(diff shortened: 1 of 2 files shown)


PatelUtkarsh:
I will update Test cases.

PatelUtkarsh:
The last `$query_post_where_actual_meta_and_snapshot_with_zero` assertion usecase was failing in previous code.

Where it was including value of SQL data which was already changed in preview. So, I use `$publish` flag to exclude it.

php/class-wp-customize-posts-preview.php
```diff
@@ -352,6 +367,23 @@ public function get_previewed_posts_for_query( WP_Query $query, $published = tru
...
+		 * @param array $post_ids Post ids being filtered.
+		 * @param array array {
+		 *     Args.
+		 *
+		 *     @type \WP_Query $query   WP_Query obj.
+		 *     @type array     $settings Field Values in snapshot.
+		 *     @type bool      $publish IN query or NOT IN query.
+		 * }
+		 */
+		$post_ids = apply_filters( 'customize_previewed_posts', $post_ids, array(
```
westonruter: I think this should be `customize_previewed_posts_for_query ` or something a bit more specific than `customize_previewed_posts` if we add a filter here.

westonruter: I think this should be `customize_previewed_posts_for_query ` or something a bit more specific than `customize_previewed_posts` if we add a filter here.

westonruter: @PatelUtkarsh And to be clear, the reason why we need a filter is because `get_previewed_posts_for_query` doesn't support all query vars, correct? If Customize Posts properly understood all query vars, then there shouldn't be a need for a new filter, right?

PatelUtkarsh: Yes, correct.
I will change filter name to `customize_previewed_posts_for_query`

PR #194, opened 2016-07-02
Add filter to return snapshot version of post_status

lgedeon:


westonruter:
This looks good! Only thing left is that there should be a unit test for this method.

westonruter:
👏

PR #195, opened 2016-07-04
Allow post field control labels to be defined in register_post_type()

lgedeon:
Check for field_x in labels when registering post_type and use as lable for default fields

### DIFF
--- a/js/customize-post-section.js
+++ b/js/customize-post-section.js
@@ -298,11 +298,12 @@
 		 */
 		addTitleControl: function() {
 			var section = this, control, setting = api( section.id );
+			var postTypeObj = api.Posts.data.postTypes[ section.params.post_type ];
 			control = new api.controlConstructor.dynamic( section.id + '[post_title]', {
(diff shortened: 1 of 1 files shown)


js/customize-post-section.js
```diff
@@ -298,11 +298,12 @@
 		 */
 		addTitleControl: function() {
 			var section = this, control, setting = api( section.id );
+			var postTypeObj = api.Posts.data.postTypes[ section.params.post_type ];
```
westonruter: @lgedeon minor thing, but in the cases where you define a new `var postTypeObj` in your patch, could you just add `postTypeObj` to the existing list of variables declared? We're generally having only one single `var` statement in our code I believe. So you could change this and other instances to:

```js
var section = this, control, setting = api( section.id ), postTypeObj;
postTypeObj = api.Posts.data.postTypes[ section.params.post_type ];
```

westonruter:
@lgedeon Thanks! 🙌

PR #196, opened 2016-07-11
Introduce Select2 to jump to the desired post in a post type panel

### LINKED ISSUE #43: Add live JS updates to post titles while waiting for selective refresh
We can do instant previews of post titles by injecting the post title into the `.entry-title` via JS as a low-fidelity preview until we wait for the selective refresh response to return. This is similar to what has been implemented for `blogname` and `blogdescription` in 4.5.

### LINKED ISSUE #124: Post List Doesn't Update in Sync with Pagination
So, I went ahead installed the plugin from the W.org repo via WPCLI. The first test was to see if the list of pagination updates in sync with the list of pagination being displayed in the theme.

Somehow, this doesn't work well. Sometimes the sticky post sticks at the top, other times it doesn't. Check this image; there are 5 posts above the posts list being displayed in the theme, those five posts are not sticky.

### LINKED ISSUE #146: List all posts of a given post type in the section
For smaller sites, where there are a limited number of total pages, it would be nice if a developer could easily customize this plugin to always list all posts of a given post type in the post type's section.

westonruter:
* Eliminate adding settings for all posts and postmeta in preview and the syncing them to the pane via postMessage. Instead only keep tally of post IDs referenced, and send them up to the pane instead. Settings are then fetched once over Ajax as needed. This greatly improves performance (if there are many posts in the preview, this can yield a 100× performance boost for loading the preview).
* Put Select2 for jumping to a post next to the Add New button, which now has its label reduced to screen reader text.
* Selecting a post in Select2 causes the section to be loaded (data fetched as needed), focused, and for the post to be navigated in the preview (if post type is `public`).
* Defer initializing partials until settings are created in preview.
* Introduce live low-fidelity JS-based previews of post title changes while waiting for selective refresh. Fixes #43.
* Let post section priority be tied to `menu_order` (if hierarchical) or `post_date_gmt` (if non-hierarchical). Fixes #124.
* Eliminate static addition of post sections in PHP, switching to only add dynamically via JS. Post sections are now always active, not hiding when the post is no longer in the preview.
* Eliminate the “no posts in preview” notice since there is now a Select2 to jump to any desired post.
* Use `customize_refresh_nonces` filter for exporting `customize-posts` nonce to ensure it is kept fresh.
* Add `wp.customize.Posts.parseSettingId()` utility function.
* Add `wp.customize.Posts.gatherFetchedPostsData()` utility function.
* Add `wp.customize.Posts.addPostSettings()` utility function.
* Add `wp.customize.Posts.ensurePosts()` utility function, allowing plugins to ensure a post is loaded and section is present.
* Add `WP_Customize_Posts::get_settings( $post_ids )`.
* Add `WP_Customize_Posts:: get_setting_params( $setting )`.

## Todo

- [x] Implement logic to fetch the post and postmeta settings for a given post ID to add a section without having to do a preview load. The pattern here has already been established in the new `insertAutoDraftPost` function which makes an Ajax request that returns a response containing all of the post/postmeta settings that would be needed for the new post section.
- [x] Potentially add a Select2 lookup field at the top of a post panel (near the new “Add Post” button) that allows you to quickly locate and load a post to edit.
- [x] Let's say you're on an archive where 10 posts are listed, and if you use the Select2 to edit a post that currently appears in the archive preview, that this should not result in navigating the preview away from the archive to the singular permalink.
- [x] Autofocus is not loading a section on demand as expected.

Fixes #124.
See #146.


westonruter:
@danielbachhuber take a look at this:

jump-to-page-1

jump-to-page-2

jump-to-page-3

Upon selection, the section is loaded, focused, and the post is loaded into the preview.

danielbachhuber:
> Upon selection, the section is loaded, focused, and the post is loaded into the preview.

Would this apply just for the content that's out of view, or does it become the new default UI for accessing any content?

My preference would be the former, because I generally agree with the UX maxim that you shouldn't hide things from users.

westonruter:
@danielbachhuber Yes, it's UI for content that is out of view. Content that is in view would get sections added automatically. In other words, if the content is not in the preview, it wouldn't get a section added, so you'd have to use a Select2 to access it unless you wanted to navigate around the site to find an instance.

ahmadawais:
UI for content that is out of view looks pretty slick @westonruter has it been pushed to the plugin?

ahmadawais:
I believe it will be in 0.7 though, checking the dev branch anyway.

westonruter:
@delawski it's in the `develop` brach, yes. Will be part of the 0.7 release.

westonruter:
@ahmadawais it's in the `develop` brach, yes. Will be part of the 0.7 release.

Issue #129, opened 2016-05-03
[possible bug] Page Editing strange behaviour


bsusala:
Tried twice, think is consistently behaving:
- open Page1
- open Editor on Page 1
- leave Editor window open
- change to another Page2 in Preview, using actual links
- in Customizer, Back
- now both pages are listed
- click on the other Page2
- Editor window still open, with text from Page1
- make any edit, Save
- Content from Page1 saves to Page2
Not good :)

westonruter:
It's hard to follow from the steps you listed. Could you create a screencast?

bsusala:
Tried again on a live, stripped down test site, the above behaviour does not replicate. Tried again on my local dev, the above reproduced until changing page inside Preview. TWO pages listed for editing, but Edit window retracts and when clicking for edit the second one, it loads the correct #2 text. Very strange. Cannot publicly post a screencast, if you want I will send a PM, no problem.

westonruter:
Are you on WordPress Slack? You can DM me a link to your screencast there.

westonruter:
@bsusala Thanks for the screencast and additional information on Slack. You noted in Slack:

> the "bug" is that if I do not save, there are two pages available for editing, which is more or less a UX "bug". not so critical, after all

And from your screencast the issue as I see it there is:

* You navigate to a page in the Customizer preview
* You open the “Pages” panel and see the one page there listed as expected.
* You open this page section to edit the page.
* You navigate to another page in the preview.
* The previous page's Customizer section is still displayed.

Issue: The Customizer section for the previous page you were editing is still displayed. If you navigate back to the Pages panel, you see two pages listed. You're saying that you expect the previous section for a page to collapse when you navigate to another page in the preview? Or rather, you expect the Customizer section for the newly-navigated page to be automatically focused when navigating to a page?

bsusala:
After reviewing the behaviour, neither :)

I am in favor of a more consistent UI. So:

CONTEXT
- editing in Customizer is, in my opinion, smth to be made after some minimal structure is in place, including static Pages
- using Customizer, in its early stage, doesn't require content editing, but more global settings
- when it's time to fine tune structure/content, navigating is more often done in Preview, rather than mingling in Customizer sections; so your Plugin should be Preview-centric, rather than Customizer-Sections-centric; all in all, your plugin is actually content focused, and content must be assessed in context (=Preview)
- links in Preview can lead to Pages, Posts or external resources, or CPTs, or other
[...]

PR #198, opened 2016-07-20
Fix editor text not persistent issue across all posts and post types

sayedwp:
Not the best solution but I could not find any other way of fixing it.  
Fix for https://github.com/xwp/wp-customize-posts/issues/129


westonruter:
@sayedwp I think I iterated onto a better solution. Please review.

sayedwp:
@westonruter Perfect! it's certainly a better a solution

PR #199, opened 2016-07-22
Include Select2 As a Dependency for Script

johnregan3:
**Adds 'select2' to the dependency array for customize-posts-panel script.**

On a fresh WP install, the customizer was not fully loading due to a JS error.

screen shot 2016-07-22 at 4 54 20 pm

Upon investigation, it was found that it was due to Select2 not being loaded into the page.


westonruter:
@johnregan3 It's good you caught this! It probably wasn't noticed previously because we've been including the [Customize Object Selector plugin](https://github.com/xwp/wp-customize-object-selector) which also includes a copy of Select2, and it probably enqueued it first.

PR #202, opened 2016-07-25
Add post date control and preview

### LINKED ISSUE #56: Implement control and preview for post_date
To start, this can be a simple text field containing the date. For specifics regarding the UI, see #53.

Care will need to be taken when modifying the date time so that the post will continue to appear in the preview when the date is in the future. Also, it would seem logical that changing the date should result in the order of the posts changing in the preview.

There may need to be a partial registered for the display of a post date. If not, then the entire content `part` will need to be refreshed (#36).

johnregan3:
Fixes #56 

## Todos
- [x] Eliminate `post_date_gmt`
- [x] Ensure `post_status` is linked properly.
- [x] Re-use `publish` status for scheduled?
- [x] Use timer to change `future` status back to `publish` when time passes?
- [x] Just update label for “Published” to “Scheduled” when date is in future?
- [x] Drafts shouldn't be assigned a date unless explicitly done so by the user? Leave empty or use placeholders when value is `0000-00-00 00:00:00`?
- [x] Fix infinite reload if modifying a partial that is not in the current view.
- [x] Sorting by previewed date?
- [x] Add unit tests.
- [x] Allow time to be reset to empty, allowing it to be set to current time upon publishing.
- [x] Handle unexpected `November 30, -0001` when post setting is invalid or empty?.
- [x] Ensure that when a setting transitions from invalid to valid, that all related post field partials are refreshed.
- [x] Add countdown timer for when post will be published?
- [x] Refine UI for post date control.
- [x] Saving a published post with current time can momentarily result in a `future` status being returned from the server before getting converted back to `publish` if the server/client times are out of sync at all. 
- [ ] ~~Date query?~~ (This will require re-architecture of `WP_Query`)
- [ ] ~~Sync post date to parent frame if opened from post preview.~~ (Do this later when status is synced.)

Update - 28 Jul 2:30 PM ET
-----
This control has been improved and refactored significantly over the last few days, especially in light of Weston's comments.

I opted to deal with the post_date specifically (as opposed to post_date_gmt) because post_date timezones are easily converted in WP using `get_gmt_from_date()`, instead of wrangling a lot of  JS.  Using this method, the user's local browser time zone is never a factor; it's has the same UX as the Post Editor.

In light of this, the JS has been greatly pared down because there isn't so much use of JS `Date` objects.

Directly updating the post_date_gmt and post_status is no longer done in this control's JS.  Now,  `WP_Customize_Post_Setting::sanitize()` handles their updates when the Preview is saved.  This functionality was almost completely in place from previous contributors.

Notes:
- A Scheduled post does not disappear from the Preview when viewing the single post, and it only disappears on an archive after the Preview is saved.
- Date inputs are now validated.
- The post modified date is handled in `sanitize()` as well.

A couple of points for improvement:
- Refactoring the JS in `js/customize-post-date.js` to eliminate using an HTML input element altogether.
- User notification:
  - Including the "Schedule for:  01 Jan 2000 @ 00:01" text above the inputs.
  - Updating the save button to "Save & Schedule" or "Save & Update" to notify the user that their post will be Scheduled, as opposed to Published.  This is similar to how the Post Edit UI currently functions.

Update - 26 Jul 11 PM ET
-----
I've got the basic functionality working as an extension of the Dynamic control.  This is found in `js/customize-posts.js`.  

I am using JS's `Date` for handling the input as it will make comparison between the existing and new dates pretty simple by using operators like "<" and ">" and "=".  This also allows us to utilize the `gmt_offset` WP option to ensure the `post_date_gmt` is truly GMT.

There are two new JS `Date` functions to handle formatting Date objects correctly based on a PHP-style format string. I went this route as they could possibly be reused elsewhere, especially when pulling in the WP options for date and time formatting.  **I may pull these out tomorrow, just to simplify things.  What thoughts do you have?**

The custom template for the inputs is found in `php/class-wp-customize-post-date-control.php`.

*The current implementation saves both `post_date` and `post_date_gmt`.*  

Up next: 

- Proper Text above the inputs (e.g., "Published on Oct 16, 2015 @ 20:31").
- Handling Post Status changes when the date is updated.
- Ensure dates are validated before updating the preview.
- Ensure posts scheduled for the future don't disappear from the preview.
- Reordering posts when date is updated.
- Need to investigate the [modified dates issue](https://github.com/xwp/wp-customize-posts/issues/56#issuecomment-222849908).

Original Text
---

I'm looking to see if I'm on the right track with my work on the Post Date UI.  This is my first time working heavily with the customizer in JS, and my first issue with Customize Posts, so this may be a bit clunky. 

My goal is to take the post_date and split it up into separate inputs just as it is in the Post Edit screen.

in the Customizer, when any input is changed, it updates a hidden field, then that value is passed over to the preview pane.  This way, we don't have to save each piece of the date (like the month, day, year, etc).


php/class-wp-customize-post-date-control.php
```diff
@@ -0,0 +1,115 @@
...
+								text = choice.text;
+								value = choice.value;
+							}
+						#>
+						<option value="{{ value }}">{{ text }}</option>
+					<# } ); #>
+				</select>
+			<# } else { #>
+				<input
+					type="text"
```
PatelUtkarsh: @johnregan3 Can we use `number` as input type with `max` and `min` attributes?

johnregan3: @PatelUtkarsh I would love to do that.  I think it would help greatly with validation.  @westonruter How closely do we want to match the existing interface?  It uses simple text inputs.

PatelUtkarsh: In number input that arrows(up and down) make it smaller so maybe that's why the core is not using it?

johnregan3: @PatelUtkarsh You know, I'm not really sure.  I think they would just rather leave it alone since it's been used for so long.  However, I think it's time for an update of that whole metabox.  in addition to the UI changes, the JS for the date function could be greatly improved.

westonruter: > How closely do we want to match the existing interface? It uses simple text inputs.

I really don't like the existing interface 😄

js/customize-post-section.js
```diff
@@ -480,6 +481,76 @@
...
+
+			/**
+			 * Split the post_date into usable parts.
+			 *
+			 * @returns {object} Object of results.
+			 */
+			function getDateInputData() {
+				var date, result = {}, singleCharLimit = 9;
+
+				date = new Date( postData.post_date );
```
westonruter: Beware of timezones.

js/customize-post-section.js
```diff
@@ -480,6 +481,76 @@
...
+					result.month = '0' + result.month;
+				}
+
+				result.day = date.getDate().toString();
+				result.year = date.getFullYear().toString();
+				result.hour = date.getHours().toString();
+				result.min = date.getMinutes().toString();
+				return result;
+			}
+			control.params.date_data = getDateInputData();
```
westonruter: This logic should be encapsulated inside of the control itself. So whenever the setting's value is updated, you can look at `control.setting().post_date` to re-parse and populate the individual inputs.

js/customize-posts.js
```diff
@@ -40,6 +41,282 @@
...
 	});
 
+	/**
+	 * Post Date control extension of Dynamic Control.
+	 *
+	 * This updates both the post_date and post_date_gmt.
+	 *
+	 * @todo update Post Status as appropriate.
+	 */
+	api.controlConstructor.post_date = api.controlConstructor.dynamic.extend({
```
westonruter: I think we'll want to put this control into its own separate JS file.

php/class-wp-customize-post-date-control.php
```diff
@@ -0,0 +1,143 @@
...
+				data-customize-setting-property-link="post_date_gmt"
+			<# } #>
+			/>
+		<input
+			id="{{ data.input_id_post_date_gmt }}"
+			type="hidden"
+			class="post-status"
+			value=""
+			<# if ( data.setting_property ) { #>
+				data-customize-setting-property-link="post_status"
```
westonruter: There is already another control for `post_status`. You can get the value via `control.setting().post_status`.

johnregan3: @westonruter Nice!  I didn't know how to fetch that directly.

js/customize-posts.js
johnregan3: Yeah.  It grew pretty large over time.  😄

westonruter:
> The current implementation saves both post_date and post_date_gmt.

Yeah, I think we should chose one. If we use `post_date` like WP does on the admin screen, then in the `WP_Customize_Post_Setting::sanitize()` function we could convert the `post_date` from the `timezone_string`/`gmt_offset` and derive the `post_date_gmt`. We could then eliminate `post_date_gmt` as being a property in the setting value entirely. Better for there to be just one then for there to be two which have to be kept in sync.

On the other hand, in other ways using `post_date_gmt` as the sole setting property makes more sense since it has an explicit timezone. If it were used, then the timezone conversion would have to be done in JS.

js/customize-post-section.js
johnregan3: @westonruter Would this go inside of `initialize()` or `_setUpSettingRootLinks()` (or somewhere else)?

js/customize-posts.js
```diff
@@ -40,6 +41,286 @@
...
+			 *
+			 *
+			 * @returns {boolean}
+			 */
+			function updatePostStatus() {
+				var gmtNow = getGmtNow();
+
+				// If the date is in the future, and it is currently published, schedule it.
+				if ( newDate > gmtNow && 'publish' === originalPostStatus ) {
+					newPostStatus.val( 'future' );
```
westonruter: This should instead set the `post_status` property on the `control.setting()` itself (via that clone/set dance).

js/customize-posts.js
```diff
@@ -40,6 +41,286 @@
...
+					currentPostStatus = 'future';
```
westonruter: You can get the current post status via `control.setting().post_status`. Shouldn't be a need for a variable.

johnregan3:
@westonruter Thanks for the feedback so far!  I think this will be much more sleek after I get all of your changes implemented.

php/class-wp-customize-post-date-control.php
```diff
@@ -0,0 +1,143 @@
...
+				data-customize-setting-property-link="post_date"
+			<# } #>
+			/>
+		<input
+			id="{{ data.input_id_post_date_gmt }}"
+			type="hidden"
+		    class="post-date-gmt"
+			value=""
+			<# if ( data.setting_property ) { #>
+				data-customize-setting-property-link="post_date_gmt"
```
westonruter: You're being resourceful by re-using the `data-customize-setting-property-link`, but I think we can be a bit more elegant by setting up our own input field data bindings.

js/customize-post-section.js
westonruter: Good question. The syncing logic should probably go into the `ready` method, that is when `control.deferred.embedded` is resolved (like you have below). You could add `control.deferred.embedded.done( ... )` to an extended `initialize` method.

westonruter: Important to note that this syncing needs to be bi-directional. So if you update the field, it should update the setting, and if the setting is updated, it should update the field. You can see this logic here: https://github.com/xwp/wp-customize-posts/blob/ffd51aef6d46bec0b45108b3f47a77f72f08f902/js/customize-dynamic-control.js#L101-L118

It's a bit more complicated in this example of a date control because you're updating _parts_ of a setting's property (like just the month or just the date).

westonruter: I could get you started with some initial code if you like.

westonruter:
> Ensure posts scheduled for the future don't disappear from the preview.

@johnregan3 This is a tricky one. In some ways, it is _desirable_ for a post to not appear in the preview because that reflects how it should appear if publishing _now_. At least, this would be the desired behavior on an archive page. Or do you mean that it disappears from the preview when viewing a singular post? It should definitely be viewable there.

johnregan3:
@westonruter Yeah.  I meant the archive.  😄

johnregan3:
@westonruter Ready for another look. 👍

westonruter:
> Updating the save button to "Save & Schedule" or "Save & Update" to notify the user that their post will be Scheduled, as opposed to Published. This is similar to how the Post Edit UI currently functions.

That's a nice thought, but since any number of other settings may also be modified beyond a post that is future-dated, the “Publish” button should remain as-is. A “Schedule” button here would only make sense in light of Customize Snapshots, see https://github.com/xwp/wp-customize-snapshots/pull/68

westonruter:
Demo video of current state: https://www.youtube.com/watch?v=Z0TuJLBiHbg

johnregan3:
@westonruter As always, checking out your improvements has been an education for me.  Nice work, buddy! 👍

westonruter:
I need some UI help. 

Here's the current date/status controls when the post is set:

screen shot 2016-07-30 at 10 00 26 am

Notice the “reset to current time” button. This will cause the `post_date` to be reset to `0000-00-00 00:00:00` ensuring that it will be set to the current time whenever the post is saved. Clicking this has the UI effect of removing the button and changing the date inputs to have placeholders which correspond to the current time (and these placeholders keep updating to reflect the current time):

screen shot 2016-07-30 at 10 00 32 am

I don't like how the description and the reset button look. They should perhaps be collapsed by default in some `details` or only appear if clicking some question mark icon or _something_. 

@johnregan3 @valendesigns @sayedwp @PatelUtkarsh @lukecarbis any ideas?

westonruter:
Also maybe the timezone info and button should appear below the inputs, while also being collapsed.

westonruter:
I've added a new scheduled countdown to appear with the timezone information when the status is `future` and the time is after now. The countdown updates dynamically each minute:

screen shot 2016-07-31 at 11 46 38 pm

If the time status is `publish` (or anything other than `future`) then the countdown does not appear:

screen shot 2016-07-31 at 11 46 58 pm

And again, you can reset the time to be empty so that it will update with the current time (notice placeholders):

screen shot 2016-07-31 at 11 46 47 pm

In these examples, notice that I've moved the description below the input fields. I'm still very much unhappy with how I've “designed” the info under the control. _Help appreciated!!_

php/class-wp-customize-post-date-control.php
```diff
@@ -0,0 +1,101 @@
...
+			<input type="number" size="4" maxlength="4" autocomplete="off" class="date-input year" data-component="year" min="1000" max="9999" />
+			@ <input type="number" size="2" maxlength="2" autocomplete="off" class="date-input hour" data-component="hour" min="0" max="23" />:<?php
+			?><input type="number" size="2" maxlength="2" autocomplete="off" class="date-input minute" data-component="minute" min="0" max="59" />
+		</div>
+		<small class="description customize-control-description">
+			<span class="scheduled-countdown"></span>
+			<?php
+			$tz_string = get_option( 'timezone_string' );
+			if ( $tz_string ) {
+				$tz = new DateTimezone( get_option( 'timezone_string' ) );
```
PatelUtkarsh: Reuse `$tz_string`

ahmadawais:
Maybe something like this would do?

ahmadawais:
@westonruter or maybe a reset icon (this one's a refresh icon though)

westonruter:
@ahmadawais great! Re: [your tweet](https://twitter.com/mrahmadawais/status/760140892778921984), I think to implement these changes, you'd add the appropriate CSS to https://github.com/xwp/wp-customize-posts/blob/ebf331ee613d1e750d9b93f6efe17782abeb3275/css/customize-posts.css#L224-L236 and move the reset button/link to be inside the `customize-control-title`: https://github.com/xwp/wp-customize-posts/blob/ebf331ee613d1e750d9b93f6efe17782abeb3275/php/class-wp-customize-post-date-control.php#L97

Open a PR to the `feature/issue-56-post-date-control` branch to amend this PR.

westonruter:
Here's where the UI has ended up with input from @ahmadawais. I've moved the timezone and schedule countdown info into a `details` element, the `summary` for which is now the timezone abbreviation:

screen shot 2016-08-01 at 8 26 56 pm
screen shot 2016-08-01 at 8 26 40 pm
screen shot 2016-08-01 at 8 25 57 pm
screen shot 2016-08-01 at 8 25 43 pm
screen shot 2016-08-01 at 8 25 33 pm

westonruter:
I'll merge this at some point over the next 12 hours unless someone finds an issue needing fixed first.

PR #206, opened 2016-08-01
Fix selecting input for newly-inserted post's first control

### LINKED ISSUE #209: Title Focus Issue While Creating New Posts
This is UX flow issue. When I click the <kbd>+</kbd> button to create a new post, the first thing I want to do is add a title. Here's where I face a UX glitch.

Current UX-Flow
- Click the <kbd>+</kbd> button
- Either press <kbd>tab</kbd> to select the title or click on the title (which BTW doesn't pre-select the existing value i.e. `(no title)`) and add a new title.

Better UX-Flow would be:
- Click the <kbd>+</kbd> button
- The title element is pre-focused, i.e. `(no title)` value is pre-selected and all I have to do is type the title.

westonruter:
This probably failed with the switch to embedding controls upon expanding a post section.

Fixes #209.

### DIFF
--- a/js/customize-post-section.js
+++ b/js/customize-post-section.js
@@ -372,7 +372,7 @@
 
 			// Select the input's contents when the value is a placeholder.
 			control.deferred.embedded.done( function() {
-				control.container.find( ':input' ).on( 'focus', function() {
+				control.container.find( 'input[type=text]' ).on( 'focus', function() {
 					if ( api.Posts.data.l10n.noTitle === control.setting().post_title ) {
 						$( this ).select();
 					}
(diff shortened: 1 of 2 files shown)


js/customize-post-section.js
```diff
@@ -341,6 +341,15 @@
 			section.postFieldControls.post_title = control;
 			api.control.add( control.id, control );
 
+			// Select the input's contents when the value is a placeholder.
+			control.deferred.embedded.done( function() {
+				control.container.find( ':input' ).on( 'focus', function() {
```
sayedwp: @westonruter What is the use of selecting all input/textarea/select/button ( `:input` ) when post title field will only be input?  I mean why not just use `input` ?

westonruter: True. It could just be `input`.

PR #211, opened 2016-08-04
Improve UX for trashing posts

### LINKED ISSUE #172: Improve UX for deleting posts

westonruter:
* Adds custom control template for post status control
* Introduces “Move to Trash” link that appears next to dropdown which appears when the status is not `trash`.
* Clicking the “Move to Trash” link sets the setting's `post_status` property to `trash` and then collapses the post section, to add a similar experience to clicking this same link on the edit post admin screen where clicking the link moves the user to the post list table. The collapse is done after a half-second delay to show the user that clicking the link just changes the dropdown, giving them a clue to restore the post by changing the status back to something other than `trash`.
* The post can be untrashed simply by re-expanding the section and changing the status to a non-trash status.

screen shot 2016-08-04 at 1 26 32 pm

screen shot 2016-08-04 at 1 26 46 pm

Fixes #172.

### DIFF
--- a/css/customize-posts.css
+++ b/css/customize-posts.css
@@ -224,11 +224,15 @@ body.customize-posts-content-editor-pane-resize #customize-preview:before {
 	text-decoration: line-through;
 }
 
-.customize-control-post_status .trash {
-	color: #a00;
+.customize-control-post_status .trash,
+.customize-control-post_status .untrash {
 	float: right;
 	display: inline-block;
 	margin-top: 5px;
+}
+
+.customize-control-post_status .trash {
+	color: #a00;
 }
 .customize-control-post_status .trash:hover {
 	color: #f00;
--- a/css/customize-posts.css
+++ b/css/customize-posts.css
@@ -220,6 +220,9 @@ body.customize-posts-content-editor-pane-resize #customize-preview:before {
 	width: auto;
 	vertical-align: middle;
 }
+.select2-results__option .trashed-title {
+	text-decoration: line-through;
+}
 
 .customize-control-post_status .trash {
 	color: #a00;
--- a/css/customize-posts.css
+++ b/css/customize-posts.css
@@ -221,6 +221,16 @@ body.customize-posts-content-editor-pane-resize #customize-preview:before {
 	vertical-align: middle;
 }
 
+.customize-control-post_status .trash {
+	color: #a00;
+	float: right;
+	display: inline-block;
+	margin-top: 5px;
+}
+.customize-control-post_status .trash:hover {
+	color: #f00;
+}
+
 .customize-control-post_date .customize-control-description {
 	font-style: normal;
 	display: block;
--- a/js/customize-post-section.js
+++ b/js/customize-post-section.js
@@ -464,12 +464,6 @@
 			section.postFieldControls.post_status = control;
 			api.control.add( control.id, control );
 
-			// Initialize the trashed UI.
-			// @todo Is this redundant with logic in post_status control's constructor?
-			api.panel( section.panel.get() ).expanded.bind( function() {
-				control.toggleTrash();
-			} );
-
 			if ( control.notifications ) {
 				control.notifications.add = section.addPostFieldControlNotification;
 				control.notifications.setting_property = control.params.setting_property;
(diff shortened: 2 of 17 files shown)


westonruter:
@lgedeon could you test please? 

@fjarrett also, could you review?

fjarrett:
@westonruter I don't see any DB write on trashed posts now. However, in my testing I tried deleted all posts/pages and created my very first page with the Customizer, then deleted it and Saved but a (no title) item kept reappearing in the list. Here's a video: https://youtu.be/pcnJTyUyd8U

westonruter:
@fjarrett thanks very much for that video!

Also, it looks like there is a problem with your Select2 dropdown, that the styles haven't been enqueued.

fjarrett:
@westonruter Running WP 4.7-alpha-38196 no plugins installed except Customize Posts, and Twenty Sixteen. Could it be because I have no posts in the database?

westonruter:
@fjarrett I don't see how. In the source, do you see a reference to `select2.css` or `select2.min.css` being enqueued and printed out? Are they 404'ing?

fjarrett:
Nothing in the console and nothing in the source.

lgedeon:
@westonruter I was able to delete, and then go back and restore a post by changing the status. Have you considered a (restore) link in place where the (move to trash) link was?

Also after deleting and going away or refreshing, I don't see a way to restore without going to the old wp-admin interface.

lgedeon:
One other observation, changing to trash manually doesn't collapse the section like it does if you click the link. I am not sure if it needs to be consistent, just noticing.

westonruter:
@lgedeon 

> Have you considered a (restore) link in place where the (move to trash) link was?

It did come to mind. But I figured for now it was just as easy to have them change status back to something else, rather than keep track of what the previous pre-trash status was (which is easy, but nonetheless). Maybe we should consider that for a future enhancement.

> Also after deleting and going away or refreshing, I don't see a way to restore without going to the old wp-admin interface.

That's true. I suppose we could address that by allowing trashed posts to appear in the Select2 dropdown results. That would be simple to do.

> One other observation, changing to trash manually doesn't collapse the section like it does if you click the link. I am not sure if it needs to be consistent, just noticing.

Yes, this was intentional.

westonruter:
@lgedeon:

> Also after deleting and going away or refreshing, I don't see a way to restore without going to the old wp-admin interface.

This has been addressed in 915a306, 1deeeb9, and 4f5cb05. You can now access trashed posts in the Select2 dropdown. Selecting one of these trashed posts causes it to be loaded _with_ the pre-trashed status and slug supplied and this new setting marked dirty from the start.

westonruter:
@lgedeon 
> It did come to mind. But I figured for now it was just as easy to have them change status back to something else, rather than keep track of what the previous pre-trash status was (which is easy, but nonetheless). Maybe we should consider that for a future enhancement.

This is added in 5bbe6d2:

screen shot 2016-08-05 at 12 52 00 am

I'm not totally sure it is really useful.

westonruter:
@fjarrett the issue with the broken Select2 style should be fixed in 58b33f4.

westonruter:
@fjarrett 
> I don't see any DB write on trashed posts now. However, in my testing I deleted all posts/pages and created my very first page with the Customizer, then deleted it and Saved but a (no title) item kept reappearing in the list. Here's a video: https://youtu.be/pcnJTyUyd8U

This issue should be fixed in https://github.com/xwp/wp-customize-posts/pull/211/commits/9852605fd8a921e454004acb95090a9c413041a8

lgedeon:
> I'm not totally sure it is really useful.
Agreed. My first thought was feature parity. I don't want to ever give anyone an excuse to exit the customizer.
That said, kids, pets, pizza sauce, accidents happen, and during that moment of panic after accidentally deleting an important post an easy obvious solution might really be appreciated.
Thank you for adding it.

lgedeon:
Looks good! 👍

fjarrett:
@westonruter Wonderful. Works great for me.

westonruter:
Thanks for testing!!

- [ ] Add unit tests.

PR #212, opened 2016-08-06
Fix for post date control breaking in firefox.

sayedwp:


westonruter:
Looking good:

image

image

PR #213, opened 2016-08-06
Line up `number` inputs in post date selection.

delawski:
@westonruter - I've fixed `number` inputs display on Firefox. Tested on Firefox, Chrome (Mac OS/Win) and IE (9, 10, 11).

I've noticed that @sayedwp worked on the same issue in #212. I had no time to look into it, but it might be a better solution than mine.

I haven't got a chance to work on `details` element. It doesn't work on only on FF, but also on IE 9, 10 and 11.

### DIFF
--- a/css/customize-posts.css
+++ b/css/customize-posts.css
@@ -264,20 +264,56 @@ body.customize-posts-content-editor-pane-resize #customize-preview:before {
 }
 .customize-control-post_date .date-inputs {
 	clear: both;
+	display: -webkit-box;
+	display: -ms-flexbox;
+	display: flex;
+	-webkit-box-align: baseline;
+	-ms-flex-align: baseline;
+	align-items: baseline;
+	-webkit-box-pack: justify;
+	-ms-flex-pack: justify;
+	justify-content: space-between;
 }
 
 .customize-control-post_date select,
 .customize-control-post_date input.date-input {
-	min-width: 10%;
-	width: auto;
+	width: 12%;
+	min-width: 12%;
+	-ms-flex-preferred-size: 8%;
+	flex-basis: 8%;
+}
+.customize-control-post_date select {
+	width: 30%;
+	-ms-flex-preferred-size: 25%;
+	flex-basis: 12%;
+	-webkit-box-flex: 6;
+	-ms-flex-positive: 6;
+	flex-grow: 6;
+}
+@media screen and ( max-width: 782px ) {
+	.customize-control-post_date select {
+		height: 40px;
+	}
 }
 .customize-control-post_date input.date-input {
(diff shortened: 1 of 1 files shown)


sayedwp:
@delawski I have already fixed both the issues, and tested that in chrome/firefox/safari, could not check in IE. 

As asked by you, I pulled your PR and got just one issue.
input-cuts-off-in-small-size
The inputs cut off in small screen.
I learned some CSS from your PR that I missed, and I am going to include those in mine.

css/customize-posts.css
```diff
@@ -264,21 +264,57 @@ body.customize-posts-content-editor-pane-resize #customize-preview:before {
...
 	clear: both;
+	display: -webkit-box;
+	display: -ms-flexbox;
+	display: flex;
+	-webkit-box-align: baseline;
+	-ms-flex-align: baseline;
+	align-items: baseline;
+	-webkit-box-pack: justify;
+	-ms-flex-pack: justify;
+	justify-content: space-between;
```
sayedwp: @delawski This will make the input fields very big in small screen, I am not sure if you wanted this.

css/customize-posts.css
```diff
@@ -264,21 +264,57 @@ body.customize-posts-content-editor-pane-resize #customize-preview:before {
...
 }
 
 .customize-control-post_date select,
 .customize-control-post_date input.date-input {
-	min-width: 10%;
-	width: auto;
+	width: 12%;
+	min-width: 12%;
```
sayedwp: If we have already set a width, giving value to `min-width` is of no use. We can reset `min-width` if required.

sayedwp:
@delawski Was thinking what would be better in small screen.

justified-content

or 

start

Justified inputs certainly look better and I like it, however since they are number fields and users are only allowed to enter 2/4 numbers, I feel it's not right to give big inputs for those fields. What's your thoughts on it?

css/customize-posts.css
sayedwp: Okay I see the reason.

westonruter:
Relevant changes have been merged into #212. Thanks!

westonruter:
> Justified inputs certainly look better and I like it, however since they are number fields and users are only allowed to enter 2/4 numbers, I feel it's not right to give big inputs for those fields. What's your thoughts on it?

That's a tough one. But yeah, probably since they should only enter 2/4 numbers (i.e. ½ 😉) into each, then maybe better to not give them the impression they can enter more.

delawski:
Good point. I think that the outer container (with `display: flex`) could have `max-width` applied to prevent justifying inputs for such edge cases.

PR #215, opened 2016-08-07
Attempt to enable translating via translate.wordpress.org

westonruter:
We're currently not able to make use of translate.wordpress.org for translating the plugin. It [reports](https://translate.wordpress.org/locale/es-mx/default/wp-plugins/customize-posts):

> This plugin is not properly prepared for localization.

projects_translated_to_spanish__mexico____glotpress_ _wordpress

I thought that the `Text Domain` plugin meta could be empty since it is now assumed to be the slug, I thought, but I added it in https://github.com/xwp/wp-customize-posts/commit/421da210a44a1b6defd7e70055d42475bed5ab9d to try to fix the issue on translate.wordpress.org. However, that didn't fix it in 0.7.0. I also thought that `Domain Path` was optional, but maybe it isn't. 

Anyone have any specifics for what is missing?

### DIFF
--- a/customize-posts.php
+++ b/customize-posts.php
@@ -8,6 +8,7 @@
  * Author URI: https://make.xwp.co/
  * License: GPLv2+
  * Text Domain: customize-posts
+ * Domain Path: /languages
  *
  * Copyright (c) 2016 XWP (https://xwp.co/)
  *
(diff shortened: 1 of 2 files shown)


westonruter:
See conversation at https://wordpress.slack.com/archives/meta-i18n/p1470963914000152

PR #216, opened 2016-08-09
Abstract Editor Control

sayedwp:


sayedwp:
@westonruter The PR is ready for review. I have tested it with multiple editor controls combinations and did not find any issue. I tried to configure `pre-commit` hook but could not do it correctly, so sorry for the failed checks.

js/customize-post-section.js
```diff
@@ -516,247 +516,18 @@
...
-			control = new api.controlConstructor.dynamic( section.id + '[post_content]', {
+			control = new api.controlConstructor.editor( section.id + '[post_content]', {
 				params: {
 					section: section.id,
-					priority: 25,
 					label: postTypeObj.labels.content_field ? postTypeObj.labels.content_field : api.Posts.data.l10n.fieldContentLabel,
-					active: true,
+					setting_property: 'post_content',
 					settings: {
 						'default': setting.id
```
westonruter: I want to see what possibilities there are for eliminating `setting_property` in favor of:

```js
settings: {
    'default': setting.id + '[post_content]'
}
``` 

That would seem more elegant, but I think it will require a core change to support.

js/customize-editor-control.js
```diff
@@ -0,0 +1,383 @@
...
+					section: '',
+					priority: 25,
+					label: api.Posts.data.l10n.fieldContentLabel,
+					active: true,
+					field_type: 'textarea',
+					setting_property: null,
+					input_attrs: {}
+				},
+				options.params || {}
+			);
```
westonruter: Todo: we need to implement the same `control.expanded`, `control.expand()`, `control.collapse()` interfaces as the widgets and nav menu items have.

js/customize-editor-control.js
```diff
@@ -0,0 +1,383 @@
...
+			 * Unlink the editor from this post and collapse the editor when the section is collapsed.
+			 */
+			section.expanded.bind( function( expanded ) {
+				if ( expanded ) {
+					api.Posts.postIdInput.val( section.params.post_id );
+				} else {
+					api.Posts.postIdInput.val( '' );
+					control.editorExpanded.set( false );
+				}
+			} );
```
westonruter: It seems that this is not enough. If there are multiple editors in a post, there seems to be a case where multiple editors can be expanded at a time.

See this new demo plugin “Customize Post Preambles”: https://gist.github.com/westonruter/f8473159e48fbacbef9dcd6498155846

If you shift-click on a preamble while the content editor is expanded, you can get into this situation:

customize__wordpress_develop_5

Notice that both of the editors have an `editorExpanded` state to true. This doesn't happen when clicking to open another editor control's “Open Editor” button however.

There seems to be a need for something like this:

```js
/**
 * Collapse any other expanded editors first.
 */
control.editorExpanded.bind( function( expanded ) {
	if ( expanded ) {
		api.control.each( function( otherControl ) {
			if ( otherControl !== control && otherControl.extended( api.EditorControl ) && otherControl.editorExpanded && otherControl.editorExpanded.get() ) {
				otherControl.editorExpanded.set( false );
			}
		} );
	}
} );
```

That being said, we also need the `expanded` state instead of `editorExpanded` as noted above, for alignment with widgets and nav menu items.

westonruter:
@sayedwp this is ready for re-review. Could you give it a try with your integration?

westonruter:
I'll also note that by implementing the expanded interface, expanding an
editor will now work seamlessly with Customizer Browser History,
automatically including the control autofocus param when the control is
expanded.

sayedwp:
@westonruter Tested with my integration , works beautifully. 👍

westonruter:
@sayedwp I have a proposal for something which could help disambiguate when there are multiple fields that could have an editor open:

customize__wordpress_develop_5

This fixes something I noticed where I would forget what the open editor was for. If you like it, would you apply some style improvements?

sayedwp:
@westonruter I was just going to say that, and I saw your commit 😄   
I had given it a thought even before, when I was going to toggle the editor animation class. I was thinking to give it tabs ( only if there are more than one editor controls) just like browser tabs. But then it will take some 50-60px of space above the editor.

westonruter:
Probably shouldn't increase the overall height any. If the title is within the bar, could maybe some gray borders on either side make it feel more natural? Maybe tabs in the future of we allow more than one editor to be expanded at once.

sayedwp:
Agree on not increasing the overall height. Will add some css to make it look natural.

sayedwp:
@westonruter Is this looking good? 

editor


The icons shows that its a standard post, as shown in post editor page


post-editor-page

westonruter:
I think the icon should be omitted because the editor can be for any post
type and for any meta now, so it may not make any sense as a thumbtack.
Looks good otherwise!

sayedwp:
Yes, didn't think of that. And that also makes me think that why you have renamed it to "Post Editor Control" from "Editor Control"? I know it would primarily be used for post content however it would also be used for meta

westonruter:
@sayedwp good point. I guess the main reason why I went with “Post” editor is that “postmeta” is still specific to posts. Also, there is a specific integration with posts in the control (the control is presumed to be added to a Post section): https://github.com/xwp/wp-customize-posts/blob/4dde895130df4c06bf072be7cdbefa41c9c745c6/js/customize-post-editor-control.js#L224-L234

westonruter:
Also, I think there should be a more generic `EditorControl` as well, eventually. Perhaps we can eventually make `PostEditControl` extend this base `EditorControl`.

sayedwp:
That would make more sense.

westonruter:
@sayedwp cool, once you've pushed up the style changes to the editor title, I think this is good to merge.

sayedwp:
@westonruter Great, lets merge it.

westonruter:
@sayedwp I did some more cleanup of the JS and added a feature where the new editor heading will only show up if there is more than one editor control in a section. If there is only the content editor control, then there would be no need to disambiguate with other controls in the section.

valendesigns:
The weird tab you guys added, which I think looks really strange sitting in the middle there, doesn't include mobile styles. So when you scale the browser down you get this.

mobile-style-bug

westonruter:
@valendesigns the tab is there to give the user context as to what they are editing. It appears when there are multiple editor controls in a given section.

@sayedwp would you update the styling to factor in mobile?

sayedwp:
Sure I will update that soon.

valendesigns:
@westonruter I understand why it was added but from a UX perspective it's misleading. It's not a tab at all. It doesn't bind events or navigate to anything. So to clear up context we've introduced confusion elsewhere. I think it would be better if we removed the tab style and just centered the text IMO. Thoughts?

westonruter:
I guess the idea is that the entire editor could be considered a tab and that's why that style was chosen. @sayedwp can you work out some non-tab styling for the heading?

PR #219, opened 2016-08-17
Issue #132 : Allow filtering to disable 'posts' component.

### LINKED ISSUE #132: Allow customize_loaded_components filter to remove posts component
Currently there is one `customize_loaded_components` filter which handles the injection of the `WP_Customize_Posts` instance into `WP_Customize_Manager`:

```php
add_filter( 'customize_loaded_components', array( $this, 'filter_customize_loaded_components' ), 100, 2 );
```

There should be another `customize_loaded_components` filter that runs earlier at priority 0 which amends `posts` to the list of components. Then the later filter can check to see of `posts` is in the array and then only instantiate `WP_Customize_Posts` if it is present.

This would allow plugins to use a more standard way to selectively disable the `posts` component.

kienstra:
@westonruter, 
Could you please review this pull request for Issue #132?

The new filter `add_posts_to_customize_loaded_components` adds `'posts'` to the array of components in the Customizer.
And `WP_Customize_Posts` is only instantiated if `'posts'` is still present.
So a later filter may remove `'posts'`, to disable it.

One issue I noticed is that if a filter does remove `'posts'`, it causes a fatal error:

```Notice: Undefined property: WP_Customize_Manager::$posts in /srv/www/wordpress-develop/src/wp-content/plugins/wp-customize-posts/php/class-customize-posts-plugin.php on line 179
Fatal error: Call to a member function add_support() on a non-object in /srv/www/wordpress-develop/src/wp-content/plugins/wp-customize-posts/php/class-customize-posts-plugin.php on line 179```

Removing the action [on line #71](https://github.com/xwp/wp-customize-posts/blob/feature/allow-removing-posts-component/php/class-customize-posts-plugin.php#L71) removed that error, but there was another fatal error.

I could have missed something here.

Fixes #132.

### DIFF
--- a/php/class-customize-posts-plugin.php
+++ b/php/class-customize-posts-plugin.php
@@ -165,6 +165,10 @@ class Customize_Posts_Plugin {
 	 */
 	function load_support_classes( $wp_customize ) {
 
+		if ( ! isset( $wp_customize->posts ) ) {
+			return;
+		}
+
 		// Theme & Plugin Support.
 		require_once dirname( __FILE__ ) . '/class-customize-posts-support.php';
 		require_once dirname( __FILE__ ) . '/class-customize-posts-plugin-support.php';
--- a/php/class-customize-posts-plugin.php
+++ b/php/class-customize-posts-plugin.php
@@ -66,6 +66,7 @@ class Customize_Posts_Plugin {
 		add_action( 'wp_default_styles', array( $this, 'register_styles' ), 11 );
 		add_action( 'init', array( $this, 'register_customize_draft' ) );
 		add_filter( 'user_has_cap', array( $this, 'grant_customize_capability' ), 10, 3 );
+		add_filter( 'customize_loaded_components', array( $this, 'add_posts_to_customize_loaded_components' ), 0, 2 );
 		add_filter( 'customize_loaded_components', array( $this, 'filter_customize_loaded_components' ), 100, 2 );
 		add_action( 'customize_register', array( $this, 'load_support_classes' ) );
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by westonruter** [comment]
@kienstra this is looking good. Regarding the fatal error, the fix should be simple. At the top of `load_support_classes` try adding:

```php
if ( ! isset( $wp_customize->posts ) ) {
    return;
}
```

That should do it. We just have to account for the same situation as another plugin that looks for whether the posts component is enabled.

kienstra:
Hi @westonruter, 
Thanks for helping with this, especially so quickly. The [commits](https://github.com/xwp/wp-customize-posts/pull/219/commits/3a905e81dcdef00c70d038696bbdcc9549a4671d) [above](https://github.com/xwp/wp-customize-posts/pull/219/commits/7dc44b870244e9660c1dce5209d018c91b8f1add) took care of errors locally, when `'posts'` is removed from `$components`.

php/class-customize-posts-plugin.php
```diff
@@ -121,17 +122,36 @@ function grant_customize_capability( $allcaps, $caps, $args ) {
...
+	 * This will allow disabling 'posts' through filtering.
 	 *
 	 * @param array                $components   Components.
 	 * @param WP_Customize_Manager $wp_customize Manager.
 	 * @return array Components.
 	 */
 	function filter_customize_loaded_components( $components, $wp_customize ) {
 		require_once dirname( __FILE__ ) . '/class-wp-customize-posts.php';
-		$wp_customize->posts = new WP_Customize_Posts( $wp_customize );
+		if ( in_array( 'posts', $components ) ) {
```
westonruter: Add a third `true` argument to `in_array()` for strict checking.

php/class-customize-posts-plugin.php
```diff
@@ -121,17 +122,36 @@ function grant_customize_capability( $allcaps, $caps, $args ) {
...
 	/**
+	 * Add 'posts' to array of components that Customizer loads.
+	 *
+	 * A later filter may remove this, to avoid loading this component.
+	 *
+	 * @param array                $components   Components.
+	 * @param WP_Customize_Manager $wp_customize Manager.
+	 * @return array Components.
+	 */
+	function add_posts_to_customize_loaded_components( $components, $wp_customize ) {
```
westonruter: Remove `$wp_customize` arg and corresponding phpdoc, as the param is not used.

kienstra:
**Thanks For Review**
_Commit To Apply Feedback_

Hi @westonruter,
Thanks for your feedback. I should have caught those points with a PHPCS check. The [commit](https://github.com/xwp/wp-customize-posts/pull/219/commits/da1f2bc8ffbfbfd52ebc7ffabab8c9cf41d569b8) above makes the corrections.

Issue #37, opened 2016-03-04
When editing a post, make sure the corresponding partial is scrolled into view


westonruter:

westonruter:
This could be implemented in the JS `Partial` instances themselves. When refreshing any partial, it could call `Element.scrollIntoView()`.

westonruter:
This is something that should be in core generally, so I've also prepared a patch for a new trac ticket I opened: https://core.trac.wordpress.org/ticket/39389

@kienstra The patch could be applied to #220 for the `deferred` partial.

PR #220, opened 2016-08-17
Issue #37 : On editing a post, scroll partial to top of preview.

kienstra:
**Request For Review**

Hi @westonruter,
Could you please review this pull request for Issue #37?

Editing a post field already triggers a refresh of its partial.
So in the `refresh()` method, define and call `scrollPartialToTopOfPage`.
Some themes have a `margin-top` of the `#page` element, like Twenty Sixteen.
And scrolling would hide part of the partial (see [screenshot](https://cldup.com/rl9EncmzzW.png) below).
So calculate the `offsetTop` of `#page`, to prevent this.
If this isn't a concern, this commit could be less verbose.

### DIFF
--- a/js/customize-deferred-partial.js
+++ b/js/customize-deferred-partial.js
@@ -66,12 +66,34 @@
 		 * @return {jQuery.Promise} Refresh promise.
 		 */
 		refresh: function() {
-			var partial = this, refreshPromise;
+			var partial = this, refreshPromise, scrollPartialToTopOfPage;
 
 			refreshPromise = api.selectiveRefresh.Partial.prototype.refresh.call( partial );
 
+			/**
+			 * Scroll the partial element to the top of the Customizer preview.
+			 *
+			 * The #page sometimes has a margin-top, like in Twenty Sixteen.
+			 * That would make this scrolling hide part of the partial element.
+			 * So get the offsetTop of #page, to account for the margin-top.
+			 * And subtract that from the scrolling value.
+			 *
+			 * @returns {void}
+			 */
+			scrollPartialToTopOfPage = function() {
+				var $partialElement, partialElementPositionY, pageElementOffsetTop;
+
+				$partialElement = document.querySelector( partial.params.selector );
+				if ( $partialElement ) {
+					partialElementPositionY = $partialElement.getBoundingClientRect().top;
+					pageElementOffsetTop = document.querySelector( '#page' ) ? document.querySelector( '#page' ).offsetTop : 0;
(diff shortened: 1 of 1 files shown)


js/customize-deferred-partial.js
```diff
@@ -66,12 +66,34 @@
...
+			 * @returns {void}
+			 */
+			scrollPartialToTopOfPage = function() {
+				var $partialElement, partialElementPositionY, pageElementOffsetTop;
+
+				$partialElement = document.querySelector( partial.params.selector );
+				if ( $partialElement ) {
+					partialElementPositionY = $partialElement.getBoundingClientRect().top;
+					pageElementOffsetTop = document.querySelector( '#page' ) ? document.querySelector( '#page' ).offsetTop : 0;
+					window.scrollBy( 0, ( partialElementPositionY - pageElementOffsetTop ) );
```
westonruter: @kienstra Yeah, there is a much simpler approach you can take here. Namely [`element.scrollIntoView()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView). If available, you could actually use [`element.scrollIntoViewIfNeeded()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoViewIfNeeded).

Something to bear in mind is whether or not these methods trigger `scroll` events. If they don't, you'll probably want to manually `trigger` it on the window so that the `scroll` message will be sent to the parent frame to remember the scroll position.

westonruter:
@kienstra Yeah, there is a much simpler approach you can take here. Namely [`element.scrollIntoView()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView). If available, you could actually use [`element.scrollIntoViewIfNeeded()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoViewIfNeeded).

Something to bear in mind is whether or not these methods trigger `scroll` events. If they don't, you'll probably want to manually `trigger` it on the window so that the `scroll` message will be sent to the parent frame to remember the scroll position.

westonruter:
@kienstra also, instead of adding this to the `refresh` method, I suggest instead that you extend the `preparePlacement` method, like so:

```js
preparePlacement: function( placement ) {
    var partial = this;
    // @todo placement.container[0].scrollIntoViewIfNeeded(); ...
    return api.selectiveRefresh.Partial.prototype.preparePlacement.call( partial, placement );
}
```

kienstra:
Thanks, @westonruter. You raise good points here.

PR #222, opened 2016-08-18
[WIP] Post editor enhancements

sayedwp:


westonruter:
FYI: There are merge conflicts.

php/class-wp-customize-post-editor-control.php
```diff
@@ -51,6 +51,94 @@ public function enqueue() {
 	}
 
 	/**
+	 * Hooked to `customize_controls_enqueue_scripts` from `WP_Customize_Posts` class.
+	 */
+	public static function enqueue_scripts() {
```
westonruter: @sayedwp Actually, maybe you should switch back to using the non-`static` `enqueue()` method, but then have a `static` class variable to capture whether the enqueue has been called:

```php
static protected $enqueued = false;

public function enqueue() {
    if ( self::$enqueued ) {
        return;
    }
    self::$enqueued = true;
    // ...
}
```

Then you can avoid `WP_Customize_Posts` from having direct awareness of this class's methods and so could remove:

```php
add_action( 'customize_controls_enqueue_scripts', array( 'WP_Customize_Post_Editor_Control', 'enqueue_scripts' ) );
```

sayedwp: @westonruter `enqueue` method never executes until we add control using `add_control`. So currently `enqueue` is just sitting there and not doing anything, we are enqueueing the script as a dependency from `register_scripts` in `Customize_Posts_Plugin`

westonruter: OK, good point. I'd consider this a deficiency in the `WP_Customize_Manager::register_control_type()` that it doesn't call the `enqueue` method.

sayedwp: I think that's because when we register the control type, it just renders the template
https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-customize-manager.php#L1517-L1543

However the actual enqueue method is called when we use `add_control` of `WP_Customize_Manager `
https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-customize-manager.php#L1633-L1636
https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-customize-manager.php#L1471-L1479

sayedwp: Also we should be able to use `customize_controls_enqueue_scripts` from inside the `WP_Customize_Post_Editor_Control` construct but we cannot do that because `render_control_template` is hooked into `customize_controls_print_footer_scripts` which is at the footer and `customize_controls_enqueue_scripts` is in the header, so already late.

https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-customize-manager.php#L301
https://github.com/WordPress/WordPress/blob/master/wp-admin/customize.php#L63-L185

However if we use `add_control` I think then it should be working because its hooked into `customize_register`.

Issue #207, opened 2016-08-01
Clear up confusion between “Save & Publish” button and publishing posts


westonruter:
If you edit a post and set its status to “draft” it is very confusing that the “Save & Publish” button mentions “Publish”. Users would assume this would switch the Draft status into Publish when you click it, but it doesn't. Via @ahmadawais:

<img width="295" alt="screen shot 2016-08-02 at 12 50 23 am" src="https://cloud.githubusercontent.com/assets/134745/17307075/27ca883a-57e8-11e6-89fc-2dda42e9ee13.png">

In the Customize Snapshots plugin, this one button is broken up into two:

<img width="299" alt="screen shot 2016-08-01 at 1 04 08 pm" src="https://cloud.githubusercontent.com/assets/134745/17307157/856a67d0-57e8-11e6-89b4-cbeee1b04e85.png">

As the Customizer state can be saved separately from making the changes live. The confusion is still present here for posts, however.

The terminology for “making changes live” needs to be improved. Instead of “Publish” it could instead be “Commit”. It could also be just “Save”, but then we'd need to change the label for snapshots to be something else.

ahmadawais:
Snapshots UI makes a lot more sense!

westonruter:
@valendesigns if we changed “Publish” (go live) to “Save”, when what do you think would be a good alternative for the current snapshot Save/Update button? I keep going back to Git terminology, like using “Commit” for publish and “Stage” or “Stash” for save/update snapshot. These aren't helpful for non-technical users.

valendesigns:
We could change "Save" to "Snapshot", but I'm not convinced that's any better or that there really is going to be anything that makes it more clear other than adding a tooltip to "Save" that explains what you're doing.

kienstra:
**Question About Solution**

Hi @westonruter,
In _Customize Posts_, would you like to change the “Save & Publish” button to instead display “Save,” as you [mentioned above](https://github.com/xwp/wp-customize-posts/issues/207#issuecomment-236800941)? And should that only be for when panels for _Customize Posts_ are open, like the “Posts" and "Pages” panels? 

It seems like it might make sense for the button to display “Save & Publish” for all other Customizer panels. Because it’ll actually publish the changes. Of course, it’s different on _Customize Snapshots_.

westonruter:
[...]

PR #223, opened 2016-08-19
Issue #207 : Rename 'Save & Publish' button to 'Save'.

kienstra:
**Request For Review**

Hi @westonruter, 
Could you please see this pull request, which sets a new name for the `'Save & Publish'` button? It's `'Save'` for the moment, pending whatever name you'd like to give it.

As you [suggested in #35547](https://core.trac.wordpress.org/ticket/35547#trac-change-1-1453324602035002), this overwrites `api.l10n.save`. When the Customizer initially loads, the value (text) of the button is `'Save & Publish'`. In about 500 milliseconds, it replaces this with `'Saved'`.

As you know, this `'Save & Publish'` text is originally produced with PHP in [customize.php#L117](https://core.trac.wordpress.org/browser/trunk/src/wp-admin/customize.php#L117). So if we'd like to prevent it from appearing at all, it would have to be overwritten with JavaScript earlier than it’s already overwritten, which I think is in [customize-controls.js#L3908](https://core.trac.wordpress.org/browser/trunk/src/wp-admin/js/customize-controls.js#L3908).

### DIFF
--- a/js/customize-posts.js
+++ b/js/customize-posts.js
@@ -588,6 +588,9 @@
 		api.previewer.bind( 'focus-control', component.focusControl );
 
 		component.ensureAutofocusConstructPosts();
+
+		// Change 'Save & Publish' button value to 'Save'
+		api.l10n.save = 'Save';
 	} );
 
 })( wp.customize, jQuery );


westonruter:
Thanks, but per https://github.com/xwp/wp-customize-posts/issues/207#issuecomment-242071767 I think we'll need to mull over this some more.

PR #225, opened 2016-08-22
Add sidebar (widget area) shortcuts control

### LINKED ISSUE #139: Add control which provides buttons to edit various sidebars
Easy access to editing the widget areas on a given template would be a key usability improvement. Currently a user has to navigate up to Widgets panel and into the widget area, and then all the way back to the page/post section after finishing. Instead the user should be able to click a button to open a given sidebar section, and then when that section is collapsed, the post section should be re-opened to return the user to where they were.

For the mechanism for using panel back button for breadcrumb navigation, see https://core.trac.wordpress.org/ticket/32683#comment:6

See also https://github.com/xwp/wp-customize-content-widgets

westonruter:
Introduction of new control when a post section which lists out all of the widget areas currently displayed in the preview. Each widget area is listed in the form of a button. When clicking this button, the widget area section is opened so that the widgets can be managed. Upon collapsing the section, the previous section is re-expanded and the widget area button originally clicked is re-focused, bringing the user back to where they were before. When the page template changes to add/remove widget areas, the widget area buttons also will show/hide. These shortcut buttons cut down a lot of clicks in the customizer UI to access the widgets and they help conceptualize widgets as content blocks.

Fixes #139.

Video demo (YouTube):

[Demo of Sidebar Shortcut Control](https://youtu.be/b1Eq2JERJC8)

### DIFF
--- a/js/customize-dynamic-control.js
+++ b/js/customize-dynamic-control.js
@@ -131,10 +131,6 @@
 			control._setUpSettingPropertyLinks();
 
 			api.Control.prototype.ready.call( control );
-
-			// @todo build out the controls for the post when Control is expanded.
-			// @todo Let the Control title include the post title.
-			control.deferred.embedded.done(function() {});
 		},
 
 		/**
--- a/js/customize-dynamic-control.js
+++ b/js/customize-dynamic-control.js
@@ -183,6 +183,28 @@
 		},
 
 		/**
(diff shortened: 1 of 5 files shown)


westonruter:
- [ ] Hide sidebar shortcuts control if there are no sidebars registered.
- [ ] Hide shortcuts control if the post is not currently being previewed?

PR #238, opened 2016-09-01
Use notification.setting instead of notification.data.setting

westonruter:
See https://core.trac.wordpress.org/ticket/37890#comment:2


westonruter:
Since `notification.data && notification.data.setting` doesn't work in core today, there's no harm in merging a check for `notification.setting` now, even though [#37890](https://core.trac.wordpress.org/ticket/37890) isn't merged.

Issue #246, opened 2016-09-04
Implement full WP_Query support


westonruter:
Ensuring that the posts returned from a `WP_Query` accurately reflect the previewed post changes is a big deficiency in the plugin currently. The current logic to handle this looks at the query vars and then injects additional `ID IN (…)` and `ID NOT IN (…)` conditions to the `WHERE` clause. See [`WP_Customize_Posts_Preview::filter_posts_where_to_include_previewed_posts()`](https://github.com/xwp/wp-customize-posts/blob/17e3cd27474cfccf5bae17bdb0424899b6b18bd9/php/class-wp-customize-posts-preview.php#L491-L517) which applies on the `posts_where` filter. That method currently has a todo:

> There are undoubtedly hundreds of possible conditions that are not being accounted for in regards to all of the possible query vars a WP_Query can take.

There is also the problem of making sure that newly-created customized posts (which have an underlying `auto-draft` state and which have `0000-00-00` as the date) appear in the proper order (the logic in [`\WP_Customize_Posts_Preview::filter_the_posts_to_preview_settings()`](https://github.com/xwp/wp-customize-posts/blob/17e3cd27474cfccf5bae17bdb0424899b6b18bd9/php/class-wp-customize-posts-preview.php#L237-L247) is insufficient because it is not aware of pagination).

I've been thinking that what we basically have to replicate the logic in `WP_Query` to augment the query results as required to accurately preview changes. Search indexes commonly intercept the query vars and add filters to get the results from another source and inject them via filtering `the_posts`. See https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-wp-query-integration.php

In WP 4.6 there is a new option and that is the new [`posts_pre_query`](https://developer.wordpress.org/reference/hooks/posts_pre_query/) filter in 4.6. (See ticket [#36687](https://core.trac.wordpress.org/ticket/36687).)

PR #247, opened 2016-09-04
Make sure that customized posts appear on the first page of results (i.e. homepage)

westonruter:
This is particularly useful when there are more than {page_per_posts} of posts on the site. If you add a new post in the customizer and then try navigating to the post index in the preview, the newly added post will not appear on the first page, even though it may have `post_date` that would put it there (even `0000-00-00` as `current_time()`, which is also why it won't appear at front of the list since it has the lowest date and so appears on last page).

The reason for needing this logic is because `WP_Query` is not fully previewable yet. See https://github.com/xwp/wp-customize-posts/issues/246


westonruter:
Closed in favor of #248.

PR #265, opened 2016-09-14
Use public instead of show_ui as default show_in_customizer value

### LINKED ISSUE #264: Restrict CPT's to public ones only
In my test environment I use WooCommerce, BBPress, and Jetpack. This might be an extreme but I also think it's a realistic prospect. Stores often have support/ community forums.

When I enable Customize Posts I see lots of CPT's for WooCommerce, BBPress, and Jetpack. There's a lot of them :)

For example in WooCommerce I see the 'Orders' CPT. But this is not something I would want to edit through the Customizer.

At the moment it looks like you're using the CPT show_ui property to determine if a post type should be listed. I've tested switching this for the public property - and it seems to tidy things up nicely. This also means you can remove the Jetpack specific code that is removing support for the Feedback CPT.

westonruter:
Fixes #264


BinaryMoon:
That looks good to me!

PR #293, opened 2016-09-23
Fix broken getCurrentTime() method in Safari 

westonruter:
In Safari, but not other browsers, `new Date( "2016-09-23 14:57:10" )` is an *Invalid Date*. It turns out that including `T` delimiting the date from the time causes it to work in Safari, so this PR implements that.

### DIFF
--- a/js/customize-posts.js
+++ b/js/customize-posts.js
@@ -533,7 +533,7 @@
 	component.getCurrentTime = function getCurrentTime() {
 		var currentDate, currentTimestamp, timestampDifferential;
 		currentTimestamp = ( new Date() ).valueOf();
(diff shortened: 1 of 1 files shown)


westonruter:
@delawski fun fact, in Chrome (and Safari) when you supply the `T` it is then the datetime string is interpreted to be in UTC. In Chrome when you don't have the `T` it interprets it in the local timezone (and Safari fails completely).

delawski:
@westonruter wow, I didn't know that making dates work in JS is still a challenge, even though it's 2016 already!

If there was a need to do complex operations on dates, I'd suggest investigating one of the date libraries like [Moment.js](http://momentjs.com/).

PR #295, opened 2016-09-24
Update customize preview post meta rows filter

PatelUtkarsh:
Added `$setting` in `customize_previewed_postmeta_rows` filter.
Added new filter `customize_previewed_postmeta_rows_{$setting->post_meta}`


PatelUtkarsh:
@westonruter before merging can you check in your local if all test-case are passing? 
Travis is passing all test case but in my VVV 3 unit-tests are breaking in develop branch without any change.

PR #309, opened 2016-10-18
Return null if can not preview customize postmeta setting

PatelUtkarsh:
Issue discussion https://github.com/xwp/wp-customize-posts/commit/0df94bb337fe091dd67ea9f4f6889dba6519ba4c#commitcomment-19463405

### DIFF
--- a/php/class-wp-customize-posts-preview.php
+++ b/php/class-wp-customize-posts-preview.php
@@ -1139,6 +1139,8 @@ final class WP_Customize_Posts_Preview {
 			);
 			if ( $can_preview ) {
 				$value = $postmeta_setting->post_value();
+			} else {
+				return null;
 			}
 
 			if ( $postmeta_setting->single ) {


PatelUtkarsh:
@westonruter not sure if this is a correct solution so feel free to amend or close this PR.

php/class-wp-customize-posts-preview.php
```diff
@@ -1139,6 +1139,8 @@ public function filter_get_post_meta_to_preview( $value, $object_id, $meta_key,
 			);
 			if ( $can_preview ) {
 				$value = $postmeta_setting->post_value();
+			} else {
+				return null;
```
westonruter: I think this makes sense, because the filter is originally invoked with `null`:

```php
$check = apply_filters( "get_{$meta_type}_metadata", null, $object_id, $meta_key, $single );
```

Also, the `null !== $value` check is done above to “Abort if another filter has already short-circuited.” So we know at this point that `$value` is `null`, so this could either return `null` like you're doing here or return `$value`.

PR #317, opened 2016-11-17
Add check for changeset status

sayedwp:


php/class-wp-customize-posts.php
```diff
@@ -701,8 +701,19 @@ public function filter_customize_save_response_for_conflicts( $response ) {
...
-		// Short circuit if there there were invalidities.
-		if ( isset( $response['setting_validities'] ) && count( array_filter( $response['setting_validities'], 'is_array' ) ) > 0 ) {
+		$has_invalidities = (
+			isset( $response['setting_validities'] )
+			&&
+			count( array_filter( $response['setting_validities'], 'is_array' ) ) > 0
+		);
+		$changeset_status_not_publish = (
+			! empty( $response['changeset_status'] )
+			&&
```
westonruter: For back-compat, I think this conditional needs to be:

```php
$changeset_status_publish = (
    empty( $response['changeset_status'] ) // Prior to 4.7, this filter only would run on actual saves.
    ||
    'publish' === $response['changeset_status']
);

// Short circuit if there there were invalidities or the changeset status was not publish.
if ( $has_invalidities || ! $changeset_status_publish ) {
```

PR #335, opened 2016-12-18
Improve logic for determining when to do fallback refresh for post field partials

westonruter:
The changes are part of some experimentation I'm doing (along with JS Widgets) to enable Customize Posts to directly mutate Backbone models created from the WP-API JS client, with the aim of demonstrating patterns for how Customize Posts can be used to manage themes that are rendered purely with JS templates. As part of this I found that we need the partials to be smarter about when they decide that they need to do the fallback behavior of a full refresh. I've introduced the idea of a “fallback dependent selector” which is used to do a test on the document to see whether or not a refresh is warranted. Namely, this selector is intended to normally match the body class or post class that would reference a given post so that if neither the current singular preview or the current archive loop mention a given post, then the fallback refresh behavior should be skipped because the element wouldn't be appearing on the page anyway. As such, this should cut down on some needless full refreshes.

* Refactor body_selector, singular_only, and archive_only into fallback_dependent_selector
* Allows partial selectors to have make use of post ID placeholders in the form of %d.
* Selective refresh fallback behavior is changed to be prevented when changing a post that isn't referenced on a given template (via body_class or post_class).
* Nevertheless, the selective refresh requests still are made so that the REST API Backbone models have the opportunity to update their rendered properties.
* Adds initial support for Twenty Seventeen.

### DIFF
--- a/js/customize-post-field-partial.js
+++ b/js/customize-post-field-partial.js
@@ -35,8 +35,6 @@
 			api.selectiveRefresh.partialConstructor.deferred.prototype.initialize.call( partial, id, args );
 
 			partial.addInstantPreviews();
-
-			// @todo If singular_only, and this is not the post singular post for this partial, then no refresh!
 		},
 
 		/**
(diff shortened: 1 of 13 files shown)


westonruter:
Please follow up on https://github.com/xwp/wp-customize-posts/pull/333

PR #341, opened 2017-01-23
Focus on specific field that have conflict

### LINKED ISSUE #35: Focus on specific post fields that have conflicts
When there are post save conflicts due to concurrent editing of a post, the post fields get an error message indicating the their version of each conflicted field. In addition to showing the error message, we should focus on the first specific conflicted field (if the section is expanded), instead of just focusing on the first control regardless.

PatelUtkarsh:
Fixes #35

### DIFF
--- a/js/customize-post-section.js
+++ b/js/customize-post-section.js
@@ -891,7 +891,8 @@
 
 			// Detect conflict errors.
 			api.bind( 'error', function( response ) {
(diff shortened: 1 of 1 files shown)


PatelUtkarsh:
Currently This PR doesn't address issue when control order is changed.

westonruter:
👏

Issue #217, opened 2016-08-15
Trashed post shows published and changes status to publish if loaded again and you hit publish.


sayedtaqui:
Step-1: Trash one post and hit Save & Publish button.
Step-2: Reload the page and search for the trashed post in dropdown and open it. Or put the section id in autofocus ( `...?autofocus[section]=sectionId` )
Step-3 Hit publish/save button again ( or go to another section make changes and hit publish )

The trashed post becomes published because the status selected in post section after loading is 'publish' and not 'trash'.

westonruter:
@sayedwp this is actually by design. When you un-trash a post it is designed to re-store the previous status and slug that were used when it was trashed previously. This follows the behavior from `wp_untrash_post()` here: https://developer.wordpress.org/reference/functions/wp_untrash_post/

You can see this behavior was specifically implemented in Customize Posts here: https://github.com/xwp/wp-customize-posts/blob/50250f165beb1192fc15aa9c5754e72b5d24d53d/php/class-wp-customize-posts.php#L1031-L1048

This being said, is there a bug here?

sayedwp:
I see a problem with this behaviour, it can make a trashed post go publish, which was not intended by user, I will create a video explaining this later today.

westonruter:
If they don't want it to be published, then they shouldn't be opening it from trash. If they want it to stay in the trash, then they should click “Move to trash” or change the status back to trash. Perhaps a notification added to the status control to alert the user that the status has been restored would be in order, and notifying them that if they publish now the post will be restored.

sayedwp:
@westonruter  Here is the video
https://www.youtube.com/watch?v=kAV-1FOtSqQ

sayedwp:
Exactly, at least adding a notification would be great, so they have some way of knowing that they are making that post published.

westonruter:
@sayedwp thanks for that helpful video. Yeah, a notification is needed here at the very least. Something like this:

<img width="300" alt="screen shot 2016-08-15 at 12 42 26 am" src="https://cloud.githubusercontent.com/assets/134745/17658528/31aa18c2-6281-11e6-8f78-c41882b211b0.png">

Example code that ads this notification:

```js
code = 'untrashed';
postStatusControlId = 'post[post][6189][post_status]';
notification = new wp.customize.Notification( code, { 
    type: 'info', 
    message: 'This has been untrashed. If you publish changes now, this post will be published. Move back to trash to avoid this.' 
} );
[...]

PR #347, opened 2017-02-27
Show notification when a post a restored from trash

sayedtaqui:
Fixes https://github.com/xwp/wp-customize-posts/issues/217

### DIFF
--- a/js/customize-posts-panel.js
+++ b/js/customize-posts-panel.js
@@ -34,6 +34,10 @@
 			panel.deferred.embedded.done(function() {
 				panel.setupPanelActions();
 			});
+
+			api.bind( 'saved', function() {
+				panel.removeTrashNotifications();
+			} );
 		},
 
 		/**
@@ -170,13 +174,13 @@
 		},
 
 		/**
-		 * Toggle trash notification if a post is restored from trash.
+		 * Toggle trash notification if a post is restored or moved to trash.
 		 *
 		 * @param {object} postData Post data.
 		 * @returns {void}
 		 */
 		toggleTrashNotification: function toggleTrashNotification( postData ) {
-			var panel = this, notification, code, statusControl;
+			var panel = this, notification, statusControl;
 
 			notification = new api.Notification( panel.trashNotificationCode, {
 				type: 'info',
@@ -188,9 +192,25 @@
 				statusControl.notifications.add( panel.trashNotificationCode, notification );
 			} );
 
-			statusControl.setting.bind( function( setting ) {
-				if ( 'trash' === setting.post_status ) {
-					statusControl.notifications.remove( panel.trashNotificationCode );
+			statusControl.setting.bind( function() {
+				statusControl.notifications.remove( panel.trashNotificationCode );
+			} );
+		},
+
+		/**
+		 * Remove trash notifications from all post sections on customize save.
+		 *
+		 * @returns {void}
+		 */
+		removeTrashNotifications: function removeTrashNotifications() {
+			var panel = this, statusControl;
+
+			api.section.each( function( section ) {
+				if ( section.extended( api.Posts.PostSection ) ) {
+					statusControl = api.control( section.id + '[post_status]' );
+					if ( statusControl && statusControl.notifications ) {
+						statusControl.notifications.remove( panel.trashNotificationCode );
+					}
 				}
 			} );
 		}
--- a/js/customize-posts-panel.js
+++ b/js/customize-posts-panel.js
@@ -19,6 +19,8 @@
 
 		postType: 'post',
 
+		trashNotificationCode: 'untrash',
+
 		ready: function() {
 			var panel = this;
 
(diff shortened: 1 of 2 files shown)


sayedtaqui:
Ready for review.

php/class-wp-customize-posts.php
```diff
@@ -846,6 +846,7 @@ public function enqueue_scripts() {
 						__( 'Customize Object Selector', 'customize-posts' )
 					)
 				),
+				'trashPostNotification' => __( 'This has been untrashed. If you publish changes now, this post will be published. Move back to trash to avoid this.' , 'customize-posts' ),
```
westonruter: It may not actually be _published_. If the post was a `draft` then it will be restored as a `draft`, right?

js/customize-posts-panel.js
```diff
@@ -32,6 +34,10 @@
 			panel.deferred.embedded.done(function() {
 				panel.setupPanelActions();
 			});
+
+			api.bind( 'saved', function() {
```
westonruter: This will trigger even when updating a changeset draft via Customize Snapshots. Is that intended? If you want it to only run when the changeset is published as a whole, then I think you need to make sure that the changeset status is `publish`, like:

```js
api.bind( 'saved', function( data ) {
    if ( 'publish' === data.changeset_status ) {
        panel.removeTrashNotifications();
    }
} );
```

This is adapted from https://github.com/xwp/wp-customize-featured-content-demo/blob/8bf4d6531adb029fbaf1e4bebb967e57989d5edd/js/featured-items-panel.js#L148-L153

Humm. But actually this code is also being used for purging the trashed items as well: https://github.com/xwp/wp-customize-posts/blob/32c6d1abcfc1b7f460603fb875110edf22d1024f/js/customize-posts.js#L1110-L1116

So maybe that other location in Customize Posts needs to be updated as well.

Or maybe this is all intended, to purge the trashed items upon saving an explicit draft as well as upon publishing, and if so you can ignore this comment 😄

php/class-wp-customize-posts.php
sayedtaqui: Correct, how about changing it to this?
"This has been untrashed. If you publish changes now, its status will change to the selected status. Move back to trash to avoid this."

js/customize-posts-panel.js
sayedtaqui: It was intended because if snapshot saves it as draft or any other status, it would eventually get published when the publish the snapshot, so the user should know this right?

php/class-wp-customize-posts.php
westonruter: Yes, that's good.

js/customize-posts-panel.js
westonruter: I think the way you have it is a better experience. If I hit Move to Trash, and then hit Save Draft, I pretty clearly don't want the trashed item around anymore and I'm unlikely to be wanting to undo it. So removing the trashed post sections upon explicit Save makes sense to me.

js/customize-posts-panel.js
```diff
@@ -160,6 +171,48 @@
...
+				message: api.Posts.data.l10n.trashPostNotification
+			} );
+
+			statusControl = api.control( postData.section.id + '[post_status]' );
+			statusControl.deferred.embedded.done( function() {
+				statusControl.notifications.add( panel.trashNotificationCode, notification );
+			} );
+
+			statusControl.setting.bind( function() {
+				statusControl.notifications.remove( panel.trashNotificationCode );
```
westonruter: So this will remove the notification whenever a user makes a change to the setting? Probably should call `statusControl.setting.unbind()` to remove this function handler after it runs once, right?

js/customize-posts-panel.js
```diff
@@ -160,6 +171,48 @@
...
+		 */
+		toggleTrashNotification: function toggleTrashNotification( postData ) {
+			var panel = this, notification, statusControl;
+
+			notification = new api.Notification( panel.trashNotificationCode, {
+				type: 'info',
+				message: api.Posts.data.l10n.trashPostNotification
+			} );
+
+			statusControl = api.control( postData.section.id + '[post_status]' );
```
westonruter: It's unlikely, but this should probably check to see if `statusControl` exists (like you're doing in `removeTrashNotifications` below), and if not then short circuit.

js/customize-posts-panel.js
```diff
@@ -160,6 +171,48 @@
...
+			statusControl.deferred.embedded.done( function() {
+				statusControl.notifications.add( panel.trashNotificationCode, notification );
+			} );
```
westonruter: Just curious why the `statusControl.deferred.embedded` is needed here. The `statusControl.notifications` collection should exist from the moment that a control is initialized, so I don't think that you have to wait for embedding?

sayedtaqui: I wasn't sure, but right embedded is not required here.

js/customize-posts-panel.js
sayedtaqui: Yes, good point

PR #348, opened 2017-02-28
Use auto-draft with late future date instead of customize-draft and clean auto-draft on changeset delete.

### LINKED ISSUE #329: Delete all customize-draft posts that are referenced by a given changeset post that is deleted
When we save a changeset (snapshot) as a `draft`, we go through and transition any related `auto-draft`-status posts to `customize-draft`. This is to prevent the posts referenced in the changeset from erroneously garbage-collected when the changeset itself is persistent. However, what we're not accounting for is the case when the changeset itself is deleted without ever having been published. What we're not doing is deleting all of these related `customize-draft`-status posts and so they are going to be sticking around forever, orphaned from any changeset.

PatelUtkarsh:
Fixes #329 

- [x] Test Use-cases. 
- [x] Write unit-test for Clean up posts on changeset delete.

### DIFF
--- a/php/class-customize-posts-plugin.php
+++ b/php/class-customize-posts-plugin.php
@@ -434,7 +434,7 @@ class Customize_Posts_Plugin {
 				}
 
 				// Confirm that this changeset is the only one which references this post so it is cleared for garbage collection.
-				$query = $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE post_type = %s AND post_status NOT IN ('trash', 'publish') AND ID != %d AND ", $post->post_type, $post->ID );
+				$query = $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE post_type = %s AND ID != %d AND ", $post->post_type, $post->ID );
 				$query .= $wpdb->prepare( 'post_content LIKE %s', '%' . $wpdb->esc_like( wp_json_encode( $setting_key ) ) . '%' );
 				$query .= ' LIMIT 1';
 				if ( $wpdb->get_var( $query ) ) { // WPCS: unprepared SQL ok.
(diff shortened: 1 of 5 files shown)


php/class-customize-posts-plugin.php
```diff
@@ -408,4 +409,39 @@ public function register_styles( WP_Styles $wp_styles ) {
...
+		$settings_data = json_decode( $post->post_content, true );
+		$status_to_delete = array( 'auto-draft', 'customize-draft' );
+		foreach ( $settings_data as $setting_key => $val ) {
+			if ( preg_match( WP_Customize_Post_Setting::SETTING_ID_PATTERN, $setting_key, $matches ) ) {
+				$setting_post = get_post( $matches['post_id'] );
+				if ( ! in_array( $setting_post->post_status, $status_to_delete, true ) ) {
+					continue;
+				}
+
+				// Confirm this is the only post setting.
```
westonruter: Maybe, “Confirm that this changeset is the only one which references this post so it is cleared for garbage collection.”

php/class-customize-posts-plugin.php
```diff
@@ -408,4 +409,39 @@ public function register_styles( WP_Styles $wp_styles ) {
...
+			return;
+		}
+		require_once( ABSPATH . WPINC . '/class-wp-customize-setting.php' );
+		require_once( dirname( __FILE__ ) . '/class-wp-customize-post-setting.php' );
+		$settings_data = json_decode( $post->post_content, true );
+		$status_to_delete = array( 'auto-draft', 'customize-draft' );
+		foreach ( $settings_data as $setting_key => $val ) {
+			if ( preg_match( WP_Customize_Post_Setting::SETTING_ID_PATTERN, $setting_key, $matches ) ) {
+				$setting_post = get_post( $matches['post_id'] );
+				if ( ! in_array( $setting_post->post_status, $status_to_delete, true ) ) {
```
westonruter: This should check if `$setting_post` is not empty, as it could have already been deleted somehow.

php/class-customize-posts-plugin.php
```diff
@@ -408,4 +409,39 @@ public function register_styles( WP_Styles $wp_styles ) {
...
+
+	/**
+	 * Delete auto-draft posts associated with the supplied changeset.
+	 *
+	 * @param int $post_id changeset/snapshot post_id.
+	 */
+	public function cleanup_autodraft_on_changeset_delete( $post_id ) {
+		global $wpdb;
+		$post = get_post( $post_id );
+		$allowed_snapshot_status = array( 'customize_changeset', 'customize_snapshot' );
```
westonruter: `$post_types`?

Issue #338, opened 2017-01-03
Create changeset prior to opening customizer for post preview


westonruter:
There is currently somewhat of a hack when opening the customizer from the post edit screen. Changes from the edit post screen get stored in `sessionStorage` and then populate the customizer settings once it is loaded. This results in an unfortunate flash of unedited content in the customizer preview and it is no longer necessary now that we have changesets (in 4.7). When clicking Preview we can now create/update a changeset with the post/postmeta changes and then open the customizer with the changeset applied.

PR #350, opened 2017-04-06
Remove sessionStorage and use changeset to save post editor screen data.

sayedtaqui:
Fixes https://github.com/xwp/wp-customize-posts/issues/338

### DIFF
--- a/js/edit-post-preview-admin.js
+++ b/js/edit-post-preview-admin.js
@@ -1,6 +1,6 @@
 /* global jQuery, _editPostPreviewAdminExports, JSON, tinymce, wp */
 /* exported EditPostPreviewAdmin */
-var EditPostPreviewAdmin = (function( $, api ) {
+var EditPostPreviewAdmin = (function( $ ) {
 	'use strict';
 
 	var component = {
@@ -39,14 +39,14 @@ var EditPostPreviewAdmin = (function( $, api ) {
 			return;
 		}
 
-		api.Loader.link = $btn;
+		wp.customize.Loader.link = $btn;
 
 		// Prevent loader from navigating to new URL.
-		wasMobile = api.Loader.settings.browser.mobile;
-		api.Loader.settings.browser.mobile = false;
+		wasMobile = wp.customize.Loader.settings.browser.mobile;
+		wp.customize.Loader.settings.browser.mobile = false;
 
 		// Override default close behavior.
-		api.Loader.close = component.closeLoader;
+		wp.customize.Loader.close = component.closeLoader;
 
 		parentId = parseInt( $( '#parent_id' ).val(), 10 );
 		if ( ! parentId ) {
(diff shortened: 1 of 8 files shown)


js/edit-post-preview-admin.js
```diff
@@ -73,11 +79,43 @@ var EditPostPreviewAdmin = (function( $ ) {
...
+			} );
+
+			request.always( function() {
+				$btn.removeClass( 'disabled' );
+				component.previewButtonSpinner.removeClass( 'is-active' );
+			} );
+		}
 
-		wp.customize.Loader.open( component.data.customize_url );
+		wp.customize.Loader.settings.browser.mobile = wasMobile;
```
westonruter: I think this needs to be put right after each of the `wp.customize.Loader.open` calls, especially the second since it is async.

php/class-edit-post-preview.php
```diff
@@ -197,4 +205,82 @@ public function make_auto_draft_status_previewable() {
...
+			wp_send_json_error( 'missing_previewed_post' );
+		} elseif ( empty( $_POST['customize_url'] ) ) {
+			status_header( 400 );
+			wp_send_json_error( 'missing_customize_url' );
+		}
+
+		global $wp_customize;
+
+		$previewed_post_id = absint( wp_unslash( $_POST['previewed_post'] ) );
+		$changeset_uuid = get_post_meta( $previewed_post_id, '_changeset_uuid', true );
```
westonruter: Maybe `_preview_changeset_uuid` would be better.

php/class-edit-post-preview.php
```diff
@@ -197,4 +205,97 @@ public function make_auto_draft_status_previewable() {
...
+		}
+
+		$customize_url = add_query_arg(
+			array(
+				'changeset_uuid' => $changeset_uuid,
+			),
+			wp_unslash( $_POST['customize_url'] )
+		);
+
+		$wp_customize_posts = new WP_Customize_Posts( $wp_customize );
```
westonruter: Wouldn't we be able to use `$wp_customize->posts` here?

sayedtaqui: Absolutely, that's what I had done first but then phpStorm was not able to recognize `$wp_customize->posts`, so.. that's probably a weird reason.

westonruter: Yeah, you're right. It's because Customize Posts is adding the `posts` property without it being declared. So I added a workaround here by adding an explicit type declaration.

php/class-edit-post-preview.php
```diff
@@ -197,4 +206,109 @@ public function make_auto_draft_status_previewable() {
...
+			wp_send_json_error( 'missing_previewed_post' );
+		} elseif ( empty( $_POST['customize_url'] ) ) {
+			status_header( 400 );
+			wp_send_json_error( 'missing_customize_url' );
+		} elseif ( empty( $_POST['input_data'] ) || ! is_array( $_POST['input_data'] ) ) {
+			status_header( 400 );
+			wp_send_json_error( 'missing_input_data' );
+		}
+
+		$previewed_post_id = intval( wp_unslash( $_POST['previewed_post'] ) );
```
westonruter: Note that this was previously `absint`. The use of this is function to sanitize integers is being discouraged in core now in favor of `intval` so that negative numbers won't get silently converted into something that could be a real post.

westonruter:
Oh, adding a unit test for `Edit_Post_Preview::update_post_changeset()` wouldn't be a bad idea.

sayedtaqui:
@westonruter Please review my changes, _( Dont have merge access )_

PR #351, opened 2017-04-19
Stop continuous ajax requests in select2

sayedtaqui:
**Issue:** When we type to select a post in post-select-lookup select2 input , it sends continuous ajax requests for each letter we type without any delay. Ideally the previous request should cancel each time a new request is sent and there should be at least some delay. This PR fixes that issue.

Example of Issue:
select2-issue

### DIFF
--- a/js/customize-posts-panel.js
+++ b/js/customize-posts-panel.js
@@ -69,7 +69,10 @@
 						});
 						request.done( success );
 						request.fail( failure );
-					}
+
+						return request;
+					},
+					delay: 250
 				},
 				templateResult: function( data ) {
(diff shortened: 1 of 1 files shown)


js/customize-posts-panel.js
```diff
@@ -69,7 +69,10 @@
 						});
 						request.done( success );
 						request.fail( failure );
-					}
+
+						return request;
```
sayedtaqui: select2 already handles aborting the request for us if we return `request` in `transport` method.

westonruter: Excellent

PR #355, opened 2017-07-23
Fix post field partial from erroneously selecting nested post placements

westonruter:
Fixes https://wordpress.org/support/topic/no-support-for-page-on-front/

### DIFF
--- a/js/customize-deferred-partial.js
+++ b/js/customize-deferred-partial.js
@@ -71,13 +71,10 @@
 		 */
 		createEditShortcutForPlacement: function( placement ) {
 			var partial = this;
-			if ( placement.container && placement.container.data( 'edit-shortcut-created' ) ) {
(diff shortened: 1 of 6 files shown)


js/customize-deferred-partial.js
```diff
@@ -61,6 +61,23 @@
...
+		 *
+		 * Prevent multiple edit shortcuts from being created on the same placement.
+		 * This should probably be upstreamed to core.
+		 *
+		 * @param {Placement} placement The placement container element.
+		 * @returns {void}
+		 */
+		createEditShortcutForPlacement: function( placement ) {
+			var partial = this;
+			if ( placement.container && placement.container.find( '> .customize-partial-edit-shortcut-button:first-child' ).length ) {
```
westonruter: The selector here wasn't right. Fixed in #356.

